PR #5, opened 2022-04-09
Porting electro-hydraulic plugin

quarkytale:
Porting plugins from bitbucket.

Install the project using instructions in [buoy_entrypoint](https://github.com/osrf/buoy_entrypoint) and run either
```
ros2 launch buoy_gazebo buoy.launch.py ign_args:="mbari_wec.sdf"
ros2 launch buoy_gazebo buoy.launch.py ign_args:="proto_electrohydraulicPTO.sdf"
```

### DIFF
--- a/buoy_gazebo/src/include/JustInterp/JustInterp.hpp
+++ b/buoy_gazebo/src/include/JustInterp/JustInterp.hpp
@@ -1,0 +1,6 @@
+#pragma once
+
+#include "JustInterp/Utils.hpp"
+#include "JustInterp/LinearInterpolator.hpp"
+#include "JustInterp/BilinearInterpolator.hpp"
+#include "JustInterp/TableInterpolator.hpp"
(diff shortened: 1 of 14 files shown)


buoy_gazebo/src/include/JustInterp/JustInterp.hpp
```diff
@@ -0,0 +1,6 @@
+#pragma once
```
chapulina, line 1: This fork is duplicate here and in `buoy_examples`:

https://github.com/osrf/buoy_examples/tree/main/include/JustInterp

It would be nice to avoid the duplication. There were a few things we had to do there to suppress linter warnings, for example. We also added a copyright header with a link to the original library.

I started a thread on Slack about this. For now, I think this PR can stay as is and we can make modifications in follow up PRs.

chapulina (approved):
I'm able to run the new world with the new plugin 👍🏽

buoy_gazebo/src/include/JustInterp/JustInterp.hpp
quarkytale: Sounds good, will create a ticket

PR #6, opened 2022-04-12
Add license

chapulina:
* `LICENSE` file
* License on both `packages.xml`s
* `CONTRIBUTING.md` file with license excerpt similar to various ROS 2 libraries (i.e. [rclc](https://github.com/ros2/rclc/blob/master/CONTRIBUTING.md)), but without the part about DCO for now
* `QUALITY_DECLARATION.md`: Updated various parts based on what's been already setup. Used [rclcpp](https://github.com/ros2/rclcpp/blob/master/rclcpp/QUALITY_DECLARATION.md) as a template.

With these changes, `colcon test` comes back clean:

```
$ colcon test --packages-select buoy_gazebo
```

```
Label Time Summary:
copyright     =   0.26 sec*proc (1 test)
flake8        =   0.23 sec*proc (1 test)
lint_cmake    =   0.16 sec*proc (1 test)
linter        =   1.09 sec*proc (5 tests)
pep257        =   0.18 sec*proc (1 test)
xmllint       =   0.26 sec*proc (1 test)
```

```
$ colcon test --packages-select buoy_description
```

```
Label Time Summary:
copyright     =   0.21 sec*proc (1 test)
lint_cmake    =   0.15 sec*proc (1 test)
linter        =   0.71 sec*proc (3 tests)
xmllint       =   0.35 sec*proc (1 test)
```

Issue #3, opened 2022-04-07
Satisfy ament linters


chapulina:
We currently have lots of ament linter failures:

```
[1.717s] The following tests FAILED:
[1.717s]  ^[[0;31m  1 - copyright (Failed)^[[0;0m
[1.717s]  ^[[0;31m  3 - cpplint (Failed)^[[0;0m
[1.717s]  ^[[0;31m  7 - uncrustify (Failed)^[[0;0m
```

We should make sure they're all passing before we turn on CI. They can be run with `colcon test` from the same directory where `colcon build` is run, and the results end up in `log/latest_test/buoy_gazebo/streams.log`

PR #7, opened 2022-04-12
Setup CI

chapulina:
Add GitHub actions-based CI that runs on every push and PR.

It uses binaries for ROS and Gazebo dependencies, pulled from https://packages.ros.org, and compiles from source all the repos in [buoy_all.yaml](https://github.com/osrf/buoy_entrypoint/blob/main/buoy_all.yaml). 

Then it tests all buoy repos, including `buoy_msgs`. I think we should just rely on the ROS buildfarm to run CI for the messages package eventually, but until we're ready to bloom it, we can test it here.

CI is currently failing the copyright check, which should be fixed by #6

### DIFF
--- a/.github/workflows/build-and-test.sh
+++ b/.github/workflows/build-and-test.sh
@@ -1,0 +1,39 @@
+#!/bin/bash
+set -ev
+
+export COLCON_WS=~/ws
+export COLCON_WS_SRC=${COLCON_WS}/src
+export DEBIAN_FRONTEND=noninteractive
+export ROS_PYTHON_VERSION=3
+
+mkdir -p $COLCON_WS_SRC
+
+apt update -qq
+apt install -qq -y lsb-release wget curl build-essential
+
+echo "deb http://packages.ros.org/ros2/ubuntu `lsb_release -cs` main" > /etc/apt/sources.list.d/ros2-latest.list
+curl -s https://raw.githubusercontent.com/ros/rosdistro/master/ros.asc | apt-key add -
+apt-get update -qq
+apt-get install -y git \
+                   python3-colcon-common-extensions \
+                   python3-rosdep \
+                   python3-vcstool \
+                   wget
+
+cd $COLCON_WS_SRC
+cp -r $GITHUB_WORKSPACE $COLCON_WS_SRC
+wget https://raw.githubusercontent.com/osrf/buoy_entrypoint/main/buoy_all.yaml
+vcs import < buoy_all.yaml
+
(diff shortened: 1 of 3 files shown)


.github/workflows/ci.yml
```diff
@@ -0,0 +1,23 @@
...
+
+jobs:
+  tests:
+    name: Build and test
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        include:
+          - gz-version: "fortress"
+            ros-distro: "galactic"
```
quarkytale, line 13: Is there supposed to be another hyphen? 
`- ros-distro: "galactic"`

.github/workflows/build-and-test.sh
```diff
@@ -0,0 +1,39 @@
+#!/bin/bash
+set -ev
```
quarkytale, line 2: Not sure about this, what's `ev`, or its `env`?

quarkytale:
Had retriggered checks, but its only failing for copyright

.github/workflows/ci.yml
chapulina: I believe it's one hyphen per matrix. See the tutorial [here](https://docs.github.com/en/github-ae@latest/actions/using-jobs/using-a-build-matrix-for-your-jobs#example-including-new-combinations). For now we only support one combination, `fortress` + `galactic`, but in the future we may want to extend to other combos (i.e. `fortress` + `humble`), and that would go on a new hyphen.

.github/workflows/build-and-test.sh
chapulina: This is actually the same as `set -e -v`. From `set --help`:

* _-e  Exit immediately if a command exits with a non-zero status._ - this is good for CI so we don't waste time trying other commands when one of them fails
* _-v  Print shell input lines as they are read._ - this is helpful when we're reading the build logs, to know what commands were running

quarkytale (approved):

chapulina:
Thanks for the review, @quarkytale ! Now that #2 and  #5 have been merged, I expect the linters to fail.

Once CI comes back, if only the linters are failing, I'll merge this PR, and you can work against `main` to address those linters as part of #3 until CI is 🟢 . If something else fails, like the build, I'll address it in this PR before merging.

chapulina:
Ok, only linter errors:

```
	  1 - copyright (Failed)
	  2 - cppcheck (Failed)
	  3 - cpplint (Failed)
```

Merging!

PR #9, opened 2022-04-13
Separate Plugin directories and removing "Proto"

quarkytale:
Created a separate directory for each plugin and their header dependencies. Added a CMakeLists function, unsure if thew include directory part is handled correctly.

Also, removed "proto" from plugin and sdf file names as well in implementation.


buoy_gazebo/CMakeLists.txt
```diff
@@ -11,25 +11,33 @@ set(IGN_GAZEBO_VER ${ignition-gazebo6_VERSION_MAJOR})
...
+    PRIVATE_LINK_LIBS
+    INCLUDE_DIRS
+  )
 
-foreach(PLUGIN ${BUOY_GZ_PLUGINS})
-  add_library(${PLUGIN} SHARED src/${PLUGIN}.cc)
-  target_link_libraries(${PLUGIN} PUBLIC
-    ignition-gazebo${IGN_GAZEBO_VER}::ignition-gazebo${IGN_GAZEBO_VER}
-    ignition-plugin${IGN_PLUGIN_VER}::ignition-plugin${IGN_PLUGIN_VER}
+  cmake_parse_arguments(gz_add_plugin "${multiValueArgs}" ${ARGN})
```
chapulina, line 23: I think this expects more arguments, we need to pass in `options` and `oneValueArgs` even if we won't use them. This worked for me:

```diff
diff --git a/buoy_gazebo/CMakeLists.txt b/buoy_gazebo/CMakeLists.txt
index 4a554cf..185f420 100644
--- a/buoy_gazebo/CMakeLists.txt
+++ b/buoy_gazebo/CMakeLists.txt
@@ -12,6 +12,8 @@ set(IGN_GAZEBO_VER ${ignition-gazebo6_VERSION_MAJOR})
 
 # Plugins
 function(gz_add_plugin plugin_name)
+  set(options)
+  set(oneValueArgs)
   set(multiValueArgs
     SOURCES
     HEADERS
@@ -20,7 +22,7 @@ function(gz_add_plugin plugin_name)
     INCLUDE_DIRS
   )
 
-  cmake_parse_arguments(gz_add_plugin "${multiValueArgs}" ${ARGN})
+  cmake_parse_arguments(gz_add_plugin "${options}" "${oneValueArgs}" "${multiValueArgs}" ${ARGN})
 
   add_library(${plugin_name} SHARED
     ${gz_add_plugin_SOURCES}
```

chapulina (commented):
Looking good so far. I believe the plugins aren't being compiled right now though. You can check this by removing your `build` and `install` directories, then compiling again, and trying to launch a world.

This is happening because we need to call `add_subdirectory` for the new `CMakeLists.txt` files to be found.

One suggestion is to add this to the top-level `CMakeLists`:

`add_subdirectory(src)`

And within `src`, you can add a `CMakeLists.txt` that has just this, so it adds all the directories within `src` by default:

```cmake
file(GLOB dirs LIST_DIRECTORIES true *)
foreach(dir ${dirs})
  if(IS_DIRECTORY ${dir})
    add_subdirectory(${dir})
  endif()
endforeach()
```

Note that this assumes that the `include` directory has been moved into one of the plugins, as suggested on another comment.

hamilton8415:
Moving the interp include files into the src directories for now as suggested sounds good.  I wasn't very happy with that library for this purpose (slow and not great in other ways...) so it could be subject to change and just placing the include files in with the source now will make that easier...  Thanks for the split out, pleased to see this coming together.

buoy_gazebo/CMakeLists.txt
```diff
@@ -11,25 +11,58 @@ set(IGN_GAZEBO_VER ${ignition-gazebo6_VERSION_MAJOR})
...
-  PolytropicPneumaticSpring
-)
+#
+# gz_add_plugin (<plugin_name>
+#              SOURCES <sources>
+#              [PUBLIC_LINK_LIBS <libraries...>]
+#              [PRIVATE_LINK_LIBS <libraries...>])
+#              [INCLUDE_DIRS <paths_to_directories...>
+#
+# Add a plugin to Gazebo GUI.
```
chapulina, line 21:
```suggestion
# Add a system plugin
```

chapulina (approved):
Well done! Just one trivial comment, but looks good to me otherwise.

PR #10, opened 2022-04-14
Satisfy ament linters

quarkytale:
Would go after #9.

Used [ament_code_style.cfg]
Copyright, cpplint and uncrustify cleared!

Flake is failing but seems unrelated:
```
4: -- run_test.py: invoking following command in '/github/home/ws/src/buoy_sim/buoy_gazebo':
4:  - /opt/ros/galactic/bin/ament_flake8 --xunit-file /github/home/ws/build/buoy_gazebo/test_results/buoy_gazebo/flake8.xunit.xml
4: 
4: ./launch/electrohydraulic_pto.launch.py:48:100: E501 line too long (116 > 99 characters)
4:             '/model/Hydraulics_Test/joint/HydraulicRam/UserCommandedCurr@std_msgs/msg/Float64]ignition.msgs.Double',

```


buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicPTO.cc
```diff
@@ -472,25 +483,36 @@ void ElectroHydraulicPTO::PreUpdate(const ignition::gazebo::UpdateInfo &_info,
   PTO_Values.I_Load = I_Load;
   PTO_Values.ScaleFactor = this->dataPtr->functor.I_Wind.ScaleFactor;
   PTO_Values.RetractFactor = this->dataPtr->functor.I_Wind.RetractFactor;
+  PTO_Values.BattChargeLimit = 0;
+  PTO_Values.BattDrawLimit = 0;
+  PTO_Values.RPMStdDev = 0;
+  PTO_Values.BiasCurrent = 0;
+  PTO_Values.Status = 0;
```
chapulina, line 490: How about defining default values in `ElectroHydraulicState.hh`? I think the linter may be happy with that, and this way it applies to every time this struct is instantiated.

chapulina (commented):
Looking good to me so far! Just a small comment.

quarkytale:
This is too messy to fix as well as review. Working on separate PRs for individual test-failures fixes.

buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicState.hh
```diff
@@ -1,55 +0,0 @@
...
-  double WindingCurrent;
-  double I_Batt;
-  double I_Load;
-  double ScaleFactor;
-  double RetractFactor;
-  double BattChargeLimit;
-  double BattDrawLimit;
-  double RPMStdDev;
-  double BiasCurrent;
-  short Status; //should be 16 bits...
```
chapulina, line 43: @andermi , was there a reason not to use `int16_t` here before? We're about to change it to `int16_t` because of this warning:

```
Use int16/int64/etc, rather than the C type short  [runtime/int] [4]
```

chapulina (commented):
> Flake is failing but seems unrelated:

Ah that was introduced by me on #13 , I'll take care of it on another PR

buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicPTO.hpp
```diff
@@ -0,0 +1,111 @@
...
+  public ignition::gazebo::ISystemPostUpdate
+{
+  /// \brief Constructor
+
+public:
+  ElectroHydraulicPTO();
+
+  /// \brief Destructor
+
+public:
```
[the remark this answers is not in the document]
quarkytale: Ahh I remember making that change in this file, must've creeped out in the mess.

chapulina (approved):
Thanks for iterating, LGTM! Let's keep the `short` change for now, let's just keep in mind that this was done in case it creates problems.

PR #11, opened 2022-04-18
Remove PneumaticSpring

chapulina:
@hamilton8415 , this is removing the prototype plugin, let me know if I'm removing too much, or forgetting to remove anything


hamilton8415 (approved):
Yep, this plugin was very basic and nothing is lost by deleting these files and making the suggested changes.

PR #12, opened 2022-04-18
Remove default plugins from SDF worlds

chapulina:
The SDF worlds are currently loading several world and GUI plugins which would be loaded by default anyway. We can reduce some boilerplate by just removing all plugins. We should only specify the plugins to load in case we want to load non-default plugins.

You can check that the world still work as before:

```
ign gazebo -v 2 mbari_wec.sdf
ign gazebo -v 2 electrohydraulicPTO.sdf
ign gazebo -v 2 pneumatic_spring.sdf
```


andermi (approved):
looks good! runs for me

PR #29, opened 2022-04-30
Fix ElectroHydraulic crash

chapulina:
A little crash was introduced in #3. Try for example:

```
ros2 launch buoy_gazebo mbari_wec.launch.py
```

Then press play, and the simulation crashes.

Originally, the `rand()` function was used, but that resulted in this warning:

```
Consider using rand_r(...) instead of rand(...) for improved thread safety.  [runtime/threadsafe_fn] [2]
```

The proposed fix here is to use a fixed a seed when calling `rand_r`. This should make the simulation more deterministic, which is usually desirable. But if we really want a different value each time the simulation is run, we could seed it with the current time.


quarkytale:
Hmm this shouldn't have creeped in, the fix for it is already merged in this [commit](https://github.com/osrf/buoy_sim/pull/10/commits/27e9f8de2b7a51496120d4855cbd2fb393a72de4)

quarkytale (approved):

PR #32, opened 2022-05-04
Adding viscous drag via hydrodynamics plugin

quarkytale:
Adding Gazebo's [Hydrodynamics](https://ignitionrobotics.org/api/gazebo/6.9/classignition_1_1gazebo_1_1systems_1_1Hydrodynamics.html) plugin on the heave cone.

Try the test world `ign gazebo src/buoy_sim/buoy_gazebo/worlds/test_heave_cone_model.sdf -v` to play with different values of plugin parameters.
___

**Parameters**
Given: Added mass in z-direction, surface area and drag coefficient.

The quadratic damping, Zww (kg/m) = 0.5 * rho * C_d * A  
where rho is density of water (kg/m^3), C_d is the drag coefficient, and A is the surface are of the cone (m^2)
A simple approach to test, used the average drag coefficient up/down to calculate Zww.
Zww(open) = 2700.10625 kg/m, Zww(close) = 3782.25 kg/m

Updated the plugin parameters <zDotW> and <zWW> with the given properties.
___

**Issues**
With the current upstream plugin, we can't change the parameters dynamically for heave cone's open/close as well as up/down motion.
___

**Discussion**
- Use separate worlds for open and close configurations
- Up and down drag coefficients: use average for now, revisit later
- May need to extend plugin in the future to add more coefficients and account for no up-down symmetry
- Write a test to drop heave cone by itself and see if it reaches the expected terminal velocity


mjcarroll:
Test in conjunction with garden's added mass.  Zero out the added mass terms in the hydrodynamics plugin but keep the viscous drag coefficients.

Does this yield something sane/stable?

quarkytale:
@hamilton8415 @andermi @mjcarroll Picking this up again. I added two buoys in a test world `heave_cone_hydrodynamics.sdf`, one having the hydrodynamics plugin and the other without it. Then plotting piston velocity from the `ElectroHydraulicPTO` plugin for both. 
- Buoy1 having the hydrodynamics plugin with just damping seems to be functioning, no crashes. Need more insights on how I can further validate damping.
- Setting this up brought up the question of bringing up multiple buoys. I did some workarounds, had to set unique namespaces for the controllers to function. And using the model name in publishers generated `ElectroHydraulicPTO` plugin so those are unique too. (Just compiling the thoughts here to be explored later).

To try this, run `ign gazebo src/buoy_sim/buoy_gazebo/worlds/heave_cone_hydrodynamics.sdf  -v 4`

https://user-images.githubusercontent.com/24695820/205039059-a48dc4dd-2adb-4249-baa3-9f8b263fff3c.mp4

quarkytale:
@andermi @hamilton8415 I'm not sure why but this is not being stable independently. Putting a pin on the sign, I can only run the sim with added mass #115 in or if I reduce the damping values from the range of 1000s to 100s.
Only test I can think of is adding an opposing force to the link externally to make it stable, but what would that imply physically?

hamilton8415:
Comments on line 485 and 498 in model.sdf.em should be , instead of Roll...

hamilton8415:
Nice to see these tests passing on CI, although I can't get the to pass locally (I never can...).  In any event, I am going to do a final review of the drag values this morning with Rob McEwen so hopefully we can merge this today if the tests really are passing.

quarkytale:
Updated with the new drag values. Are we defaulting the heave cone to be doors closed? Also, need advice on whether to keep the test world for testing values in future?

hamilton8415 (approved):
I checked the sdf values and they match the documentation.  I also built and ran this and the buoy behaved as one would expect.  I am not able to get the tests to pass locally, although I see they are fine on the CI.  This seems to be a problem for me locally recently, if they pass for you locally as well as on CI, merge away.

andermi:
@quarkytale @hamilton8415 is there anything left to do here or can I merge this so I can merge batch running?

hamilton8415:
I say merge and move on.  We will come back after the waves are merged as we've discussed.

Issue #31, opened 2022-05-04
Add physical properties to mbari_wec model


andermi:
I've put properties from Fran&ccedil;ois in a spreadsheet. These should be added to the mbari_wec model and/or accounted for in plugins.
[mbari_wec_properties.xlsx](https://github.com/osrf/buoy_sim/files/8626607/mbari_wec_properties.xlsx)

quarkytale:
Thanks for compiling and sharing this!
Making sure if I'm reading the values correctly, the added mass (closed/open) is in z-direction right? Also could you please help me identify the individual terms of Inertia (kg*m^2).

chapulina:
I'm going to work on the Inertia, weight and CoG properties for all links in a standalone PR now. It would help me with #26 so I can work with more realistic dynamics while updating the cable. And it should help your buoyancy efforts too, @quarkytale .

> could you please help me identify the individual terms of Inertia (kg*m^2). I'm assuming its the 3x3 rotational inertia matrix with the format:

I'm going with this same assumption and will verify visually if it looks reasonable.

andermi:
> the added mass (closed/open) is in z-direction

correct!

> 3x3 rotational inertia matrix with the format:

also correct!

chapulina:
Some values we're currently estimating but could need more precise numbers in the future:

- Piston moments of inertia
- Have cone inertia without Trefoil
- Trefoil inertia

chapulina:
For record-keeping, we've moved the CoB of the buoy a bit up to `2.46 m` instead of `2.16 m` in #33 so the box volume matches the shape of the buoy a bit better. We may want to tweak this in the future.

Issue #16, opened 2022-04-19
Add buoyancy plugin


chapulina:
Add the [Buoyancy](https://ignitionrobotics.org/api/gazebo/6.9/classignition_1_1gazebo_1_1systems_1_1Buoyancy.html) plugin to the example worlds and make sure it behaves reasonably. 

Upstream's buoyancy plugin supports simple horizontal surface, so it will probably be more appropriate for sub-surface links.

chapulina:
> Upstream's buoyancy plugin supports simple horizontal surface, so it will probably be more appropriate for sub-surface links.

For the initial implementation, let's use this plugin for the buoy link as well. We need to make sure we set the appropriate [`density_change`](https://gazebosim.org/api/gazebo/6.9/classignition_1_1gazebo_1_1systems_1_1Buoyancy.html#details) at the surface of the water, and that the buoy is using a supported collision shape (box or sphere).

If that approximation isn't enough, then we look into implementing a better free surface buoyancy model (#19 ).

chapulina:
Some possible improvements for the existing buoyancy plugin:

* Support attaching it to either the world or individual models
* Support passing in the volume for each link instead of relying on the collision shape

PR #33, opened 2022-05-05
Adding buoyancy

quarkytale:
Adding Gazebo's [Buoyancy](https://gazebosim.org/api/gazebo/6.6/classignition_1_1gazebo_1_1systems_1_1Buoyancy.html#details) plugin to the world.

PTO and Heave Cone volume should come to `0.3139 m^3` and `0.1928 m^3`, respectively. Assuming buoy volume to be `~5.47 m^3` (using 1m height).
Also, updated center of buoyancy from [#31](https://github.com/osrf/buoy_sim/issues/31). Since collision won't be used for anything else right now, made heave cone and buoy collision geometry to be a cuboid, pto is using mesh.
- Graded buoyancy with water density `1025 km/m^3`
- Added water plane
- Removed WaterPlanarArea joint, buoy seems stable

 

To run: 
```
ign gazebo mbari_wec.sdf
```

### DIFF
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -72,11 +72,11 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
         </material>
       </visual>
       <collision name="collision">
-        <pose>0 0 2.16 0 0 0 </pose>
+        <pose>0 0 2.26 0 0 0 </pose>
         <geometry>
-          <mesh>
-            <uri>meshes/buoy_float.stl</uri>
-          </mesh>
+          <box>
+            <size>2.65 2.65 4.65</size>
+          </box>
         </geometry>
       </collision>
     </link>
@@ -108,7 +108,6 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
         </material>
       </visual>
       <collision name="collision">
-        <pose>0 0 -3.04 0 0 0 </pose>
         <geometry>
           <mesh>
             <uri>meshes/pto_collision.stl</uri>
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -72,10 +72,10 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
         </material>
       </visual>
       <collision name="collision">
-        <pose>0 0 2.26 0 0 0 </pose>
+        <pose>0 0 2.46 0 0 0 </pose>
         <geometry>
           <box>
-            <size>2.65 2.65 4.65</size>
+            <size>2.34 2.34 1</size>
           </box>
         </geometry>
       </collision>
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -267,9 +267,9 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
       <collision name="HeaveConeCollision">
         <pose>0 0 -1.21 0 0 0 </pose>
         <geometry>
-          <mesh>
-            <uri>meshes/heave_cone.stl</uri>
-          </mesh>
+          <box>
+            <size>0.578 0.578 0.5771</size>
+          </box>
         </geometry>
       </collision>
     </link>
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -71,6 +71,14 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
           <specular>1.0 1.0 0.0 1</specular>
         </material>
       </visual>
+      <collision name="collision">
+        <pose>0 0 2.16 0 0 0 </pose>
+        <geometry>
+          <mesh>
+            <uri>meshes/buoy_float.stl</uri>
+          </mesh>
+        </geometry>
+      </collision>
     </link>
 
     <link name="PTO">
@@ -100,6 +108,7 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
         </material>
       </visual>
       <collision name="collision">
+        <pose>0 0 -3.04 0 0 0 </pose>
         <geometry>
           <mesh>
             <uri>meshes/pto_collision.stl</uri>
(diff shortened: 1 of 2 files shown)


buoy_description/models/mbari_wec/model.sdf.em
```diff
@@ -264,6 +264,14 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
           <specular>0.1 0.1 .1 1</specular>
         </material>
       </visual>
+      <collision name="HeaveConeCollision">
+        <pose>0 0 -1.21 0 0 0 </pose>
+        <geometry>
+          <mesh>
+            <uri>meshes/heave_cone.stl</uri>
+          </mesh>
```
chapulina, line 272: Using this mesh to get the volume, it's being automatically calculated to almost 8 m^3:

image

According to @andermi on Slack, that should be closer to 0.1928 m^3.

I think it's best that we approximate the collision using a box whose dimensions result in the desired volume.

chapulina (commented):
Just some quick feedback

buoy_description/models/mbari_wec/model.sdf.em
andermi: do you need collision meshes for all links?

quarkytale: Added a collision box of size `0.578 0.578 0.5771` which makes the volume to `0.1928 m^3`

quarkytale: @andermi I'm guessing collision meshes for all links would be good in the long run. I have a question though from `pto_collision.stl` the volume is coming to `0.161425` but in the Slack thread it was supposed to come out to `0.3139 m^3`. Should I use an approximate box too for the PTO or the mesh is more accurate?

chapulina: The collisions are having 2 separate roles at the moment:

* Actually work as collisions that bump into each other
* Defining the volume that's used by the buoyancy plugin

That duality may become cumbersome for links like the PTO, which need both properties, with different shapes.

Possible solutions:

1. Quickest: 
    * Leave PTO buoyancy coming from the STL for now
1. Almost as quick: 
    * Use proper STL meshes for what needs to collide, and boxes for buoyancy
    * The PTO would have 2 collisons with different roles
    * Exclude the PTO STL collision from the buoyancy plugin using the `<enable>` tag to explicitly enable all other collisions
    * Make sure the PTO's buoyancy collision can't physically collide with the tether
1. Best
    * Improve the buoyancy plugin with the suggestions in https://github.com/osrf/buoy_sim/issues/16#issuecomment-1115403060

I think we could start with `1.` on this PR and work on `3.` next.

buoy_description/models/mbari_wec/model.sdf.em
```diff
@@ -100,6 +108,7 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
         </material>
       </visual>
       <collision name="collision">
+        <pose>0 0 -3.04 0 0 0 </pose>
```
chapulina, line 111: We shouldn't move this collision, it's actually being used as a collision so we keep the tether in place. This is what it looks like with this PR:

image

buoy_description/models/mbari_wec/model.sdf.em
```diff
@@ -71,6 +71,14 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
           <specular>1.0 1.0 0.0 1</specular>
         </material>
       </visual>
+      <collision name="collision">
+        <pose>0 0 2.16 0 0 0 </pose>
```
chapulina, line 75: This one too, if we're using the STL mesh, it can't have an offset:

image

buoy_description/models/mbari_wec/model.sdf.em
```diff
@@ -71,6 +71,14 @@ tether_bottom_link_cylinder.mass_matrix(tether_bottom_link_mm)
...
+        <geometry>
+          <mesh>
+            <uri>meshes/buoy_float.stl</uri>
```
chapulina, line 78: tl;dr: we should use a box here

The automatically calculated volume for this mesh is 17.0038 m^3.

The spec sheet says the water plane area is 5.47 m^2. If we consider 17 m^3 to be the correct submergible volume, that would mean that the buoy has > 3 m of submergible height, which seems too much for me. That's my way of saying that I think 17 is too high.

But even if 17 were the correct volume, an STL mesh wouldn't work well when we add a water plane using `<graded_buoyancy>`. So we need to use a box collision.

The dimensions of the box collision should be such that `dX * dY == 5.47` (the water plane area). I'm not sure about the height though. Do we know the submerged volume of the buoy at rest, @andermi ? We should be able to calculate `dZ` with that.

chapulina (changes requested):
The heave cone collision looks good to me. I have comments on the others.

chapulina:
Another thing I should have mentioned before is that the ultimate test for the buoyancy should be to remove the joint that's currently preventing the model from sinking.

The buoyancy added in this PR should be tuned so that when the joint is removed:

* the model doesn't sink - if it does, we'll need to add more collisions or increase the ones we have
* the model doesn't fly away - we'll need the `graded_buoyancy` for that, so the buoy stops floating when it reaches the plane between the water and the air.

Judging by past experience, getting to the right balance may be very tricky and we may need to iterate on it every time something changes in the model. Let's see how the first pass goes and adjust as we go.

buoy_description/models/mbari_wec/model.sdf.em
quarkytale: Not sure how much of the buoy should be covered for collision.
When I aimed for a volume of 5.47 (given 1m height), it just covers slightly more than half. Note the pose had to shifted too, the sheet had center of volume to be `0 0 2.16`.
half_collision_buoy

And this is the params with a box fully covering it.
full_collision_buoy

How should I proceed?

quarkytale:
Sneak peek of buoy submerged :)
water_plane

buoy_description/models/mbari_wec/model.sdf.em
chapulina: Let's go with the one on the top, the collision with only 1m height. That should match the displace water volume better.

chapulina: > Note the pose had to shifted too, the sheet had center of volume to be 0 0 2.16.

Ah I had missed that before. The current position looks good to me, because we're just approximating the actual volume at the moment. I think we can get this first iteration in with some offset and adjust as we add other forces.

chapulina (approved):
Great to see the joint to the world gone! And nice touch with the water plane 💯 

I shook the buoy a bit and it seems to be moving reasonably! No longer constrained to a single axis 😄 

buoy_buoy

PR #28, opened 2022-04-29
Add spring bridge and split launch files

chapulina:
* Builds on top of #27 
* Follow up to #13 

---

### Summary

* Move the PTO bridge into its own launch file
* Add a bridge launch file for the topics exposed by the spring plugin
* Use these launch files conditionally from the entrypoint `mbari_wec.launch.py`
* Update `electrohydraulicPTO.launch.py` to be just a wrapper around `mbari_wec` 
* Clean the linters so we can finally have 🟢 CI

### Test it

See the arguments provided by the main entrypoint launch file:

<details><summary>ros2 launch buoy_gazebo mbari_wec.launch.py --show-args</summary>
<p>

```
Arguments (pass arguments as '<name>:=<value>'):

    'ign_args':
        Simulator arguments
        (default: 'mbari_wec.sdf')

    'model_name':
        Name of model
        (default: 'MBARI_WEC')

    'has_pto':
        Whether to bridge the ElectroHydraulicPTO.
        (default: 'True')

    'pto_joint_name':
        Name of PTO prismatic joint
        (default: 'HydraulicRam')

    'has_lower_spring':
        Whether to bridge the lower PolytropicPneumaticSpring.
        (default: 'true')

    'lower_spring':
        Name of lower chamber for polytropic pneumatic spring
        (default: 'lower_polytropic')

    'has_upper_spring':
        Whether to bridge the upper PolytropicPneumaticSpring.
        (default: 'true')

    'upper_spring':
        Name of upper chamber for polytropic pneumatic spring
        (default: 'upper_polytropic')
```

</p>
</details>

Test the different worlds

* `mbari_wec`
    1. Launch the default `mbari_wec` world and bridge all its plugins:
        `ros2 launch buoy_gazebo mbari_wec.launch.py`
    1. Press play
    1. Try out some ROS topics, i.e.
        `ros2 topic echo /force_lower_polytropic`
        `ros2 topic echo /battcurr_HydraulicRam`

* `electrohydraulicPTO`
    1. Launch the `electrohydraulicPTO` test world and bridge all its plugins:
        `ros2 launch buoy_gazebo electrohydraulic_pto.launch.py`
    1. Press play
    1. Try out some ROS topics, i.e.
        `ros2 topic echo /force_lower_polytropic`
        `ros2 topic echo /pistonvel_HydraulicRam`

PR #34, opened 2022-05-05
Clear up flake8

chapulina:
I was originally doing this through #28, but since that PR is on hold, here's a more direct fix.

Let's have 🟢 CI!


quarkytale (approved):
Works for me, tests cleared!

PR #35, opened 2022-05-06
Update inertias

chapulina:
* Part of #31 

Updated the inertial properties according to the values provided.

## Buoy

buoy_inertia

## PTO

* Visualization has a bug that's being fixed here: https://github.com/ignitionrobotics/ign-math/pull/424

pto_inertia_1

## Piston

We don't have the inertial terms yet, so I just eyeballed them for now

piston_inertia

## Heave cone

I subtracted 10 kg from the mass to compensate for the Trefoil, but didn't change the moments for now.

heave_inertia

## Trefoil

Kept the mass at 10 kg and eyeballed the inertia for now

trefoil_inertia

### DIFF
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -22,6 +22,10 @@ tether_top_length = 2.0
 
 num_tether_bottom_links = 10
 
+# Heave cone
+heave_total_mass = 817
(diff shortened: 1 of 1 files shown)


quarkytale (approved):
Looks good! Verified with buoyancy as well, works.
(Piston and Trefoil inertias pending)

PR #36, opened 2022-05-06
Move joint to world out of model

chapulina:
The `WaterPlaneArea` joint is just a temporary trick to prevent the model from sinking. We'll remove it eventually, but until then, it would be convenient to move it out of the model, so that the model can be reused in other worlds without that trick.

I actually did this as part of #26 so that the playground world can have a different joint to the world. Moving it to a separate PR for quicker review, and to remove future conflicts when the joint is removed.

This change shouldn't change the behaviour of the `mbari_wec` world.


quarkytale (approved):
Works for me!

PR #40, opened 2022-05-11
Move world plugins to a config file

chapulina:
On #12 all world plugins were removed from the world files so we could load the defaults instead. But then on #33 we needed a non-default plugin, which meant we had to add all others.

This PR moves those plugins to a `server.config` file and points to that file using the `IGN_GAZEBO_SERVER_CONFIG_PATH`, so that becomes the new default set of plugins. This way, buoyancy will be enabled by default for all worlds.

It's still possible to override this config file by explicitly listing plugins inside each world file as needed.

There's more info about server configs on [this tutorial](https://gazebosim.org/api/gazebo/6.9/server_config.html).

## Test it

Run

`ign gazebo mbari_wec.sdf -r`

The simulation should start and the buoy bounces up and down due to buoyancy.

### DIFF
--- a/buoy_gazebo/CMakeLists.txt
+++ b/buoy_gazebo/CMakeLists.txt
@@ -68,6 +68,7 @@ add_subdirectory(src)
 install(DIRECTORY
   worlds
   launch
+  gazebo
   DESTINATION share/${PROJECT_NAME}/
 )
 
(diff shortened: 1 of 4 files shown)


quarkytale (approved):
Works for me!

Issue #30, opened 2022-05-02
Aggregate data in ECM and use system to publish to ROS 2


chapulina:
As an alternative to publishing all values as doubles and bridging each to a ROS topic, the approach here would be to put the values in a custom component, and then have a single system that's ROS-enabled and publishes a message with all of them.

andermi:
@hamilton8415 to assist with any of these; I can make the first Controller as an example for the others. @rhenthorn 
 let me know if this looks good to you.
@chapulina should we open a separate issue for commands these controllers would need to handle?

We can break these into separate PR's that make sense

Todo:
- [ ] @andermi Add SCRecord data to ECM from PneumaticSpring
- [ ] @rhenthorn Add PCRecord, BCRecord data to ECM from ElectroHydraulicPTO
- [ ] @rhenthorn Add TFRecord data to ECM
- [ ] @andermi Add XBRecord data to ECM
- [ ] @andermi Publish SCRecord data from ECM in SpringController ROS2 System Plugin
- [ ] @rhenthorn Publish PCRecord data from ECM in PowerController ROS2 System Plugin
- [ ] @rhenthorn Publish BCRecord data from ECM in BatteryController ROS2 System Plugin
- [ ] @rhenthorn Publish TFRecord data from ECM in TrefoilController ROS2 System Plugin
- [ ] @andermi Publish XBRecord data from ECM in CrossbowController ROS2 System Plugin

andermi:
@hamilton8415 @rhenthorn Can you provide a description of the status fields for the __Records? (or point me to the docs for them)

hamilton8415:
My understanding is you're asking for the bit-field descriptions in the status words from the various converters.  It's a closely held secret!!  ;)  

Kidding of course, a good start is from the source code for each controller, copied in here and generally in "types.h" in each of the controller firmwares on bitbucket:

SC_
struct StatusBits {
    unsigned int ReliefValveRequest : 7;  //Request to open/close valve
    unsigned int ReliefValveStatus : 1;      // Status of Relief valve open/close
    unsigned int PumpRequest : 1;   // Request to turn pump on or off
    unsigned int PumpStatus : 1;  // Status of pump switch
    unsigned int PumpOverTemp : 1;  // Status of pump OverTemp signal
    unsigned int PumpToggle : 1;  // Status of pump Toggle.
    unsigned int TetherPowerRequest : 1;  //Request to turn tether power on or off
[...]

PR #41, opened 2022-05-12
add XBow AHRS ROS2 plugin

andermi:
addresses part of #30 

- [x] @andermi Add XBRecord data to ECM from plugin or use sensor
- [x] @andermi Publish XBRecord data from ECM in CrossbowController ROS2 System Plugin

passes `colcon test`


chapulina (approved):
LGTM 👍🏽

PR #52, opened 2022-05-31
Setup profiler

chapulina:
[Ignition Common's profiler](https://ignitionrobotics.org/api/common/4.4/profiler.html) is very handy to debug performance.

This PR just sets up the profiler and makes use of the existing profiling points.

See the snippet added to the README for how to enable and use it. In order to get profiling points coming from the Gazebo libraries themselves, they need to be compiled from source with `-DENABLE_PROFILER=1`.

Here's some example output. I was surprised to see that the `PolytropicPneumaticSpring` is taking up almost as much time as the physics' forward step.

buoy_profiler

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,3 +2,16 @@ This repository contains packages used for Gazebo simulation of marine buoys.
 
 Start from [Buoy entrypoint](https://github.com/osrf/buoy_entrypoint).
 
+## Development
+
+### Profiler
+
+By default, the profiler isn't enabled. To enable it at compilation time,
+pass the `-DENABLE_PROFILER=1` CMake argument. For example, to compile tests
+and enable the profiler with `colcon`:
+
+```
+colcon build --cmake-args ' -DBUILD_TESTING=false' ' -DENABLE_PROFILER=1'
+```
+
+See more on [this tutorial](https://ignitionrobotics.org/api/common/4.4/profiler.html).
--- a/buoy_gazebo/CMakeLists.txt
+++ b/buoy_gazebo/CMakeLists.txt
@@ -1,6 +1,15 @@
 cmake_minimum_required(VERSION 3.5)
 project(buoy_gazebo)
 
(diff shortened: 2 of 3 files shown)


hamilton8415:
This is great, glad to see such a tool exists.  I was surprised that the ElectroHydraulicPTO plugin didn't win the award.  I know for sure there's some optimizations in there still to chase.

andermi:
> the `PolytropicPneumaticSpring` is taking up almost as much time as the physics' forward step

Most likely my debug console printouts. I can remove most of them now.

andermi:
@chapulina when I try this, I don't see SimulationRunner in the Remotery window

Screenshot from 2022-06-02 17-03-23

chapulina:
> when I try this, I don't see SimulationRunner in the Remotery window

Did you compile all of Ignition Fortress from source with `-DENABLE_PROFILER=1`? If you haven't, you'll only see the profiling points from `buoy_gazebo`.

Other tips to navigate the UI may also help here. Use the scroll wheel to zoom in a bit more time-wise, and click on the `+` button a few times next to `Thread0` to see more lines. The `Pause` button on the top-right also helps.

andermi:
> Did you compile all of Ignition Fortress from source

Ah, that's the issue. Thanks!

andermi:
There, I fixed it! :1st_place_medal: 

Screenshot from 2022-06-06 18-15-59

andermi (approved):
:+1: 

It also might be worth mentioning that I had to run:
`colcon build --cmake-args ' -DBUILD_TESTING=true' ' -DENABLE_PROFILER=0'`
in my workspace to get `colcon test` to work again (even on other branches... since it's in the cmake in `build` in top level workspace)

chapulina:
> It also might be worth mentioning that I had to run:

Ahh good call, it took me a while to figure that out the first time I needed it, it's definitely worth mentioning.

Issue #44, opened 2022-05-16
Command services in ROS2 Spring Controller plugin


andermi:
Companion to #30.
Spring Controller should accept ROS2 commands via services and communicate commands to Pneumatic Spring plugins

Todo:
~~- [x] `SCPackRateCommand` via `/sc_pack_rate_command`~~
- [x] sc pack rate via parameter
- [x] `ValveCommand` via `/valve_command`
- [x] `PumpCommand` via `/pump_command`
- [ ] `SCResetCommand` via `/sc_reset_command`
- [ ] `BenderCommand` via `/bender_command`
- [ ] `TetherCommand` via `/tether_command`

andermi:
@hamilton8415 Should we call this done? I don't think there's much to model with the Bender (that's the ground fault monitor?), Tether (that just turns on/off the power to the heave cone?), or SC Reset (just reboots the uC?).

hamilton8415:
Yes, The "bender" command is actually in the battery controller, but does turn on and off the ground fault detection as you say.  All we need to model in this environment for the spring controller is "valve", "pump", "sc_packrate".  "Tether" turns on and off the heave cone controller.  We could have the spring controller accept this and have a way to signal the electro-hydraulic PTO plugin (which could represent the added current draw), and the trefoil controller (which would go silent if off).  Or we could have those pieces of code respond to a "tether on/off" command directly.  Just a choice, probably no wrong answer.  A "SC_Reset" command has probably never been issued at sea or testing (that's some damn solid firmware! ;)    Practically, it would just reset the packet sequence number to zero without changing much of anything else  (although it would turn the tether power on as that happens when the controller starts).  I would say it doesn't need to be modeled.

andermi:
closing the rest of these until/if we decide we need them

PR #53, opened 2022-06-02
Add pump to SpringController / Pneumatic Spring plugin

andermi:
Addresses #44 Pump command

Test with:
`ros2 service call /pump_command buoy_msgs/srv/PumpCommand duration_sec:\ 10`

Can turn off pump with:
`ros2 service call /pump_command buoy_msgs/srv/PumpCommand duration_sec:\ 0`

piston retracts ~2 inches per minute

- [x] Add command service to SpringController
- [x] Model pump in Pneumatic Spring plugin

### DIFF
--- a/buoy_gazebo/src/PolytropicPneumaticSpring/PolytropicPneumaticSpring.cpp
+++ b/buoy_gazebo/src/PolytropicPneumaticSpring/PolytropicPneumaticSpring.cpp
@@ -205,7 +205,7 @@ void PolytropicPneumaticSpring::pumpOn(
   double mass_flow = this->dataPtr->pump_absement * this->dataPtr->pump_pressure;
   double delta_mass = dt_sec * mass_flow;  // kg of gas per step
 
-  ignerr << "pump -- delta mass from/to this chamber to/from other: " << delta_mass << " kg" <<
+  igndbg << "pump -- delta mass from/to this chamber to/from other: " << delta_mass << " kg" <<
     std::endl;
 
   IGN_ASSERT(V1 >= this->dataPtr->dead_volume, "volume of chamber must be >= dead volume");
--- a/buoy_gazebo/src/PolytropicPneumaticSpring/PolytropicPneumaticSpring.cpp
+++ b/buoy_gazebo/src/PolytropicPneumaticSpring/PolytropicPneumaticSpring.cpp
@@ -69,6 +69,12 @@ struct PolytropicPneumaticSpringPrivate
   /// \brief measure of valve opening cross-section and duration (meter-seconds)
   double valve_absement{49e-7};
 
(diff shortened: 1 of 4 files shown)


andermi:
@hamilton8415 I wasn't sure how much pressure the pump creates (I can see in some of the schematics, it's a linear actuator pump like a bicycle pump), so I picked a value above what I think the max pressure would be in the lower chamber, and then tuned the retraction to 2 inches per minute using the absement.

buoy_gazebo/src/PolytropicPneumaticSpring/PolytropicPneumaticSpring.cpp
```diff
@@ -211,7 +260,9 @@ void PolytropicPneumaticSpring::Configure(
   igndbg << "is upper? " << std::boolalpha << this->dataPtr->is_upper << std::noboolalpha <<
     std::endl;
 
-  this->dataPtr->valve_absement = SdfParamDouble(_sdf, "valve_absement", 5e-6);
+  this->dataPtr->valve_absement = SdfParamDouble(_sdf, "valve_absement", 49e-7);
```
chapulina, line 263: Can you pass `this->dataPtr->valve_absement` here to avoid repeating the default value? The same for other lines

```suggestion
  this->dataPtr->valve_absement = SdfParamDouble(_sdf, "valve_absement", this->dataPtr->valve_absement);
```

chapulina (approved):
Works for me. I noticed that the piston velocity seems to be considerably higher than `2 in / min` (`0.0008 m / s`) though. 

image

buoy_gazebo/src/PolytropicPneumaticSpring/PolytropicPneumaticSpring.cpp
andermi: I'm fixing all of this in #55

buoy_gazebo/src/controllers/SpringController/SpringController.cpp
```diff
@@ -254,8 +304,7 @@ struct SpringControllerPrivate
           ros_->node_->get_logger(),
           "Pump off after (" << \
             services_->command_watch_.ElapsedRunTime().seconds() << "s)");
-        igndbg << "piston moved: " << \
-        (state.range_finder - init_x) / \
+        igndbg << "piston moved: " << (state.range_finder - init_x) /
```
andermi, line 307: > velocity seems to be considerably higher

Rather than rely on instantaneous velocity (since we have a newtonian nightmare of springs/dampers bobbing around), I have been computing the average velocity as piston position (stop-start) / pump duration

andermi: that said, something seems to have changed and made the mean velocity higher...

Issue #45, opened 2022-05-16
Command services in ROS2 Power Controller plugin


andermi:
Companion to #30.
Power Controller should accept ROS2 commands via services and communicate commands to ElectroHydraulicPTO plugin.

Todo:
- [ ] `PCBattSwitchCommand` via `/pc_batt_switch_command`
- [ ] `PCBiasCurrCommand` via `/pc_bias_curr_command`
- [ ] `PCChargeCurrLimCommand` via `/pc_charge_curr_lim_command`
- [ ] `PCDrawCurrLimCommand` via `/pc_draw_curr_lim_command`
- [ ] `PCPackRateCommand` via `/pc_pack_rate_command`
- [ ] `PCRetractCommand` via `/pc_retract_command`
- [ ] `PCScaleCommand` via `/pc_scale_command`
- [ ] `PCStdDevTargCommand` via `/pc_std_dev_targ_command`
- [ ] `PCVTargMaxCommand` via `/pc_v_targ_max_command`
- [ ] `PCWindCurrCommand` via `/pc_wind_curr_command`
- [ ] `GainCommand` via `/gain_command`

andermi:
I believe most of these can be done as parameters except `PCBattSwitchCommand`

PR #57, opened 2022-06-08
add PowerController ROS2 node

andermi:
Addresses #30 #45

Power Controller should accept ROS2 commands via services and communicate commands to ElectroHydraulicPTO plugin.

Feedback -- Todo:
- [x] Add `PCRecord` data to ECM from ElectroHydraulicPTO
- [x] Publish `PCRecord` data from ECM in PowerController ROS2 System Plugin
~~- [ ] Implement PowerController Status bit-field~~

Command -- Todo:
- [x] `publish_rate` via parameter
- [x] `PCWindCurrCommand` via `/pc_wind_curr_command`
- [x] `PCScaleCommand` via `/pc_scale_command`
- [x] `PCRetractCommand` via `/pc_retract_command`
- [x] `PCBiasCurrCommand` via `/pc_bias_curr_command`
~~- [ ] `PCBattSwitchCommand` via `/pc_batt_switch_command`~~
~~- [ ] `PCChargeCurrLimCommand` via `/pc_charge_curr_lim_command`~~
~~- [ ] `PCDrawCurrLimCommand` via `/pc_draw_curr_lim_command`~~
~~- [ ] `PCStdDevTargCommand` via `/pc_std_dev_targ_command`~~
~~- [ ] `PCVTargMaxCommand` via `/pc_v_targ_max_command`~~
~~- [ ] `GainCommand` via `/gain_command`~~
~~- [x] `PCPackRateCommand` via `/pc_pack_rate_command`~~

### DIFF
--- a/buoy_gazebo/CMakeLists.txt
+++ b/buoy_gazebo/CMakeLists.txt
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(buoy_gazebo)
 
+# if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+#   add_compile_options(-Wall -Wextra -Wpedantic)
+# endif()
+
 # Option to enable profiler
 option(ENABLE_PROFILER "Enable Ignition Profiler" FALSE)
 
@@ -30,6 +34,7 @@ set(GZ_SIM_VER ${ignition-gazebo6_VERSION_MAJOR})
 #              [PUBLIC_LINK_LIBS <libraries...>]
 #              [PRIVATE_LINK_LIBS <libraries...>])
 #              [INCLUDE_DIRS <paths_to_directories...>
+#              [ROS]
 #
 # Add a system plugin
 #
@@ -44,6 +49,8 @@ set(GZ_SIM_VER ${ignition-gazebo6_VERSION_MAJOR})
 #
 # [INCLUDE_DIRS]: Specify a list of path to directories to be included,
 #
+# [ROS]: Signal that this plugin is ROS-enabled
+
 function(gz_add_plugin plugin_name)
   set(options ROS)
   set(oneValueArgs)
@@ -73,6 +80,7 @@ function(gz_add_plugin plugin_name)
   endif()
   target_include_directories(${plugin_name}
       PUBLIC ${gz_add_plugin_INCLUDE_DIRS})
+  # target_compile_features(${plugin_name} PUBLIC c_std_99 cxx_std_17)  # Require C99 and C++17
   install(
     TARGETS ${plugin_name}
     DESTINATION lib)
--- a/buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
+++ b/buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
@@ -48,7 +48,6 @@
 class WindingCurrentTarget
 {
 public:
-  // static constexpr double DEFAULT_SCALE_FACTOR;
   double TorqueConstantNMPerAmp;   // N-m/Amp
   double TorqueConstantInLbPerAmp;  // in-lb/Amp
   JustInterp::LinearInterpolator<double> DefaultDamping;
--- a/buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
+++ b/buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
@@ -48,6 +48,7 @@
 class WindingCurrentTarget
 {
 public:
+  // static constexpr double DEFAULT_SCALE_FACTOR;
   double TorqueConstantNMPerAmp;   // N-m/Amp
   double TorqueConstantInLbPerAmp;  // in-lb/Amp
   JustInterp::LinearInterpolator<double> DefaultDamping;
@@ -68,7 +69,7 @@ public:
   /// \brief mutex to protect jointVelCmd
   std::mutex UserCommandMutex;
 
-  WindingCurrentTarget(void)
+  WindingCurrentTarget()
   {
     std::vector<double> N {0.0, 300.0, 600.0, 1000.0, 1700.0, 4400.0, 6790.0};       // RPM
     std::vector<double> Torque {0.0, 0.0, 0.8, 2.9, 5.6, 9.8, 16.6};       // N-m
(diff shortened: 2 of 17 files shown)


andermi:
@chapulina any tips for cleaning up this commit history? I've always been shy of rebase...

andermi:
@hamilton8415

https://github.com/osrf/buoy_sim/blob/e3a63531572981e2e5c4bd61bc97f0076371acf5/buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicPTO.cpp#L353

Is this supposed to be proportional to x? Wondering if it's stuck at 12.12 from debugging previously

andermi:
Status bit-field will be implemented in another PR. Other commands may be implemented in future PR.

chapulina:
> I was thinking these messy commits/merges might make it harder to view the diffs for this specific PR

Some reviewers like to go commit-by-commit. Personally, I think only specific types of PRs can benefit from that. I tend to do my reviews file-by-file, I think GitHub's UI really helps with that.

So, no need to change for me 😄

andermi:
https://github.com/osrf/buoy_sim/blob/e3a63531572981e2e5c4bd61bc97f0076371acf5/buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicPTO.cpp#L353

> Is this supposed to be proportional to x? Wondering if it's stuck at 12.12 from debugging previously

@hamilton8415 bump

buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
```diff
@@ -48,6 +48,7 @@
 class WindingCurrentTarget
 {
 public:
+  // static constexpr double DEFAULT_SCALE_FACTOR;
```
chapulina, line 51: Are you planning to use this?

buoy_gazebo/src/buoy_utils/CommandTriState.hpp
```diff
@@ -0,0 +1,87 @@
...
+// Unless required by applicable law or agreed to in writing, software
+// distributed under the License is distributed on an "AS IS" BASIS,
+// WITHOUT WARRANTIES OR CONDITIONS OF ANY KIND, either express or implied.
+// See the License for the specific language governing permissions and
+// limitations under the License.
+
+#ifndef BUOY_UTILS__COMMANDTRISTATE_HPP_
+#define BUOY_UTILS__COMMANDTRISTATE_HPP_
+
+#include <iostream>
```
chapulina, line 18: I think this include isn't used

buoy_gazebo/CMakeLists.txt
```diff
@@ -1,6 +1,10 @@
 cmake_minimum_required(VERSION 3.5)
 project(buoy_gazebo)
 
+# if(CMAKE_COMPILER_IS_GNUCXX OR CMAKE_CXX_COMPILER_ID MATCHES "Clang")
+#   add_compile_options(-Wall -Wextra -Wpedantic)
```
chapulina, line 5: Are you planning to enable these? I can help addressing the rising issues if needed

chapulina (commented):
Did an initial high-level review. It looks to me that some parts of this PR could be broken off into separate PRs so they're merged faster while you wait for feedback on specific points.

buoy_gazebo/CMakeLists.txt
andermi: I noticed they are enabled in ROS2, but we haven't been using them for the buoy repos. Is this something OpenRobotics doesn't typically worry about for Gazebo?
Also:
`target_compile_features(${plugin_name} PUBLIC c_std_99 cxx_std_17)`

I thought ROS2 needed c++17, but the plugin seems to compile without this enabled.

buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
andermi: I was thinking about getting rid of all the `#define` for constants. I decided to do it in a cleanup PR.

buoy_gazebo/CMakeLists.txt
chapulina: For the Gazebo libraries themselves, we enable the flags [here](https://github.com/gazebosim/gz-cmake/blob/035016c9900ce3628c93ca025f5d6bd5f6c41ccb/cmake/IgnSetCompilerFlags.cmake#L127-L131).

Good point that this is enabled for core ROS 2 packages, I think it's a good idea to enable them here if we want to target a high quality level for this package 👍🏽 

As for C++17, I think we may be inheriting `CMAKE_CXX_STANDARD`, but maybe there's something else going on.

buoy_gazebo/src/controllers/PowerController/PowerController.cpp
```diff
@@ -0,0 +1,779 @@
...
+
+          this->dataPtr->ros_->pc_pub_->publish(pc_record);
+        } else {
+          data.unlock();
+        }
+
+        this->dataPtr->ros_->pub_rate_->sleep();
+      }
+    };
+  this->dataPtr->thread_publish_ = std::thread(publish);
```
chapulina, line 670: I know this is following the pattern from other controllers that have already been merged, but did you consider publishing directly from the `PostUpdate` function instead of spinning a separate thread? It may simplify the logic that's copying and protecting variables.

chapulina (commented):
The code looks good to me in general, but I'm not sure how to manually test this.

When I send each command, what change in behaviour should I expect to see?

btw, we really need to start adding docs 😄

andermi:
For now, I've been checking that the overridden value gets set in the `/pc_record`. The effects of bias, scale, retract, and winding current modify the damping in the piston motion. Accounting for the sign of these values, they should resist piston velocity. Winding current is calculated proportional to motor rpm by default. Users may do some other sensible thing. Scale is factor is applied to the winding current. Setting a large scale factor should damp out motion faster than low scale. Positive bias should result in retracting the position, negative bias should extend. Retract factor is used multiplicatively with scale in the retraction direction only.

chapulina (approved):
LGTM

What I checked manually, for reference:

1. `ros2 launch buoy_gazebo mbari_wec.launch.py`
2. `ros2 topic echo /pc_record`
3. `ros2 service call /pc_wind_curr_command buoy_msgs/srv/PCWindCurrCommand wind_curr:\ 10.0\`
4. Check that echoed `wcurrent` changes to ~10 and then back to ~0
5. `ros2 service call /pc_retract_command buoy_msgs/srv/PCRetractCommand retract:\ 1.0\`
6. Check that echoed `retract` value goes to 1.0
7. `ros2 service call /pc_scale_command buoy_msgs/srv/PCScaleCommand scale:\ 0.0\`
8. Check that echoed `scale` value goes down to 0.5 (but not to 0.0)
9. `ros2 service call /pc_bias_curr_command buoy_msgs/srv/PCBiasCurrCommand bias_curr:\ 10.0\ `
10. Check that echoed `bias_current` goes to 10

Issue #58, opened 2022-06-08
Implement status bit-field for SpringController


andermi:
from `types.h` in SpringController Firmware:
```
struct StatusBits {
    unsigned int ReliefValveRequest : 7;  //Request to open/close valve
    unsigned int ReliefValveStatus : 1;      // Status of Relief valve open/close
    unsigned int PumpRequest : 1;   // Request to turn pump on or off
    unsigned int PumpStatus : 1;  // Status of pump switch
    unsigned int PumpOverTemp : 1;  // Status of pump OverTemp signal
    unsigned int PumpToggle : 1;  // Status of pump Toggle.
    unsigned int TetherPowerRequest : 1;  //Request to turn tether power on or off
    unsigned int TetherPowerStatus : 1;  //Status of tether power relay
    unsigned int LR_Fault : 1;  //Status of LRF fault input
    unsigned int AUX_Fault : 1;  //Status of AUX fault input
};

typedef  union
{
    unsigned int status;
    struct StatusBits bits;
}  STATUS;
```

_Originally posted by @hamilton8415 in https://github.com/osrf/buoy_sim/issues/30#issuecomment-1149426524_

hamilton8415:
Here are the four status bits that have anything to do with the Pump in the SpringController firmware, more detailed description below
    unsigned int PumpRequest : 1;   // Request to turn pump on or off
    unsigned int PumpStatus : 1;  // Status of pump switch
    unsigned int PumpOverTemp : 1;  // Status of pump OverTemp signal
    unsigned int PumpToggle : 1;  // Status of pump Toggle.

PumpRequest:  This is a bit that gets set when a CAN command requests the pump is run.  Set to one or zero on such a command.  This also gets set to zero if an assocaited "pump_ctr" variable counts down to zero.

PumpStatus:  This represents the actual state of the power-supply that turns the pump power on.  Can differ from PumpRequest briefly after request, and before conditions are right to turn on the power supply.

PumpOverTemp:  The pump itself is a packaged servo-motor and lead-screw arrangement that cycles a small piston pump (Cazenave custom design!  His first design at MBARI :) back and fourth.   That servo-motor controller has a digital output that indicates over-heating.  If this is set (it's never happened), the pumping pauses until that bit turns off and the status is reported here.  I doubt the pump timer stops during this pause, so if the pump ever fails to produce a pressure change, this would be a place to look.

[...]

PR #63, opened 2022-06-16
add status bit struct for SpringController SCRecord

andermi:
Addresses #58


chapulina (approved):
Nothing stood out to me as an issue. But I'm not familiar with the system and how it should behave. Maybe we should start adding some simple tests for the plugins, they usually work as documentation for how the system should behave too. Let me know if I can help getting that rolling.

andermi:
Tests for the `status` field in `SCRecord` have been added and pass in #70

PR #64, opened 2022-06-21
Publish tf_static with transient_local durability

chapulina:
* Targets #50 
* Requires https://github.com/gazebosim/ros_gz/pull/259

## Tracking down the issue

@quarkytale , I figured out what the `tf_static` issue was. It turns out it's a QoS mismatch. I found out because when I tried to view TF with

```
ros2 run tf2_tools view_frames
```

I'd see a message saying 

```
[WARN] [1655778779.765880584] [view_frames]: New publisher discovered on topic '/tf_static', offering incompatible QoS. No messages will be received from it. Last incompatible policy: DURABILITY
```

Then with the following, I saw that the subscriber was expecting `TRANSIENT_LOCAL` durability (i.e. latched message)

```
ros2 topic info /tf_static -v
```

## Changes

* Override the durability for the `tf_static` topic using a parameter on the launch file. That requires #259 to be merged upstream.
* Remove the static transform that was filling the gap so far
* Depend on `ros_ign_bridge` - that helps when building in a colcon workspace, and should also help CI / releasing
* Reduce the static publish frequency to 1 Hz (it defaults to 50Hz), because that doesn't change anyway
* Use the model frame `MBARI_WEC` instead of the canonical link frame `MBARI_WEC/Buoy` for odom. That's the proper frame to use, because in the future we may add an offset between these 2.

### DIFF
--- a/buoy_gazebo/launch/mbari_wec.launch.py
+++ b/buoy_gazebo/launch/mbari_wec.launch.py
@@ -84,15 +84,8 @@ def generate_launch_description():
             (link_pose_gz_topic, '/tf'),
             (link_pose_gz_topic + '_static', '/tf_static'),
         ],
+        parameters=[{'qos_overrides./tf_static.publisher.durability' : 'transient_local'}],
         output='screen'
-    )
-
-    # Mapping initial frame
-    tf_buoy = Node(
-        package='tf2_ros',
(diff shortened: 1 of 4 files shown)


quarkytale (approved):
Thanks for looking into this!

PR #65, opened 2022-06-24
Removing colcon warning

quarkytale:
This should take care of the warning while building `buoy_tests` package.

```
[75.327s] WARNING:colcon.colcon_cmake.task.cmake.build:Could not run installation step for package 'buoy_tests' because it has no 'install' target
```

Modified the `BUILD_TESTING` flag logic. Ideally, `return()` should bring back control to the parent `CMakeLists.txt`, but here there is none which might be causing the issue.

### DIFF
--- a/.github/workflows/build-and-test.sh
+++ b/.github/workflows/build-and-test.sh
@@ -33,6 +33,7 @@ rosdep install --from-paths ./ -i -y -r --rosdistro $ROS_DISTRO
 source /opt/ros/$ROS_DISTRO/setup.bash
 cd $COLCON_WS
 colcon build --packages-up-to buoy_tests --event-handlers console_direct+
+source $COLCON_WS/install/setup.bash
 
 # Test all buoy packages
 colcon test --packages-select-regex=buoy --event-handlers console_direct+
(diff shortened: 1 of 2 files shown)


quarkytale:
Tests are passing on my system after sourcing the workspace

.github/workflows/build-and-test.sh
```diff
@@ -33,6 +33,7 @@ rosdep install --from-paths ./ -i -y -r --rosdistro $ROS_DISTRO
 source /opt/ros/$ROS_DISTRO/setup.bash
 cd $COLCON_WS
 colcon build --packages-up-to buoy_tests --event-handlers console_direct+
+source $COLCON_WS/install/setup.bash
```
chapulina, line 36: I'm curious, is this needed? I thought `colcon test` could run without explicitly sourcing

chapulina (approved):
Thanks, @quarkytale!

.github/workflows/build-and-test.sh
quarkytale: CI was throwing errors without it!

quarkytale: Had seen the same on my system too, sourcing cleared it

chapulina: Ahh I saw what's happening now, LGTM!

PR #70, opened 2022-07-06
Add Spring Controller ROS command tests

andermi:
Valve and Pump tests

Commit history is a bit of a mess, but new files are:

- `sc_commands_ros_feedback.cpp`
- `sc_commands_ros_feedback.launch.py`

Also, adjusted `pump_absement` in `mbari_wec.sdf`

- [x] needs `NOT_IMPLEMENTED` flag from https://github.com/osrf/buoy_msgs/pull/13#
- [x] needs status field flags from https://github.com/osrf/buoy_msgs/pull/16#


quarkytale (approved):

andermi:
These tests pass every time on my machine, but they seem to do some odd things in CI. Could one of you try in your local workspace?

quarkytale:
Tests failing for me too, even with the recent commit. Both `launch_no_inputs_py.launch.py` and `launch_sc_commands_ros_feedback_py.launch.py` failed with the same error:

`rosidl_generator_py.import_type_support_impl.UnsupportedTypeSupport: Could not import 'rosidl_typesupport_c' for package 'buoy_msgs'`

I guess it might be due to that flake error, checking if it goes away with the fix

andermi:
It's gotta be a command timing issue... I'll try on another machine.

andermi:
my jetson nano reproduces the issue! My edit/build/test cycle is like 5 times as long now though lol

chapulina:
Doubling the timeout seemed to be enough, see how `launch_sc_commands_ros_feedback.launch.py` passed here for [this commit](https://github.com/osrf/buoy_sim/commit/5ac49e499f01f5eb362d73d65fa5dceead09962f):

https://github.com/osrf/buoy_sim/runs/7398781130?check_suite_focus=true

Ideally our tests wouldn't be taking so long, but I think this can be revisited later. For now maybe that's enough to unblock this PR?

andermi:
> Ideally our tests wouldn't be taking so long

Yeah, for the command tests, the test is spending a bit of time waiting for the buoy to settle. We could just spawn the buoy in a stable spot to cut out the first 15sec of sim time of each test.

quarkytale:
It doesn't seem the test will finish even with increasing the timeout to >2 hours, it kept on running for me till I aborted it. Same with CI.
```
[Processing: buoy_tests]                                       
Aborted  <<< buoy_tests [1h 53min 45s]   
```

andermi:
I have it working on my fast desktop computer, and my super slow, low-memory Jetson Nano. I don't know what is hanging on CI.

quarkytale:
I can help with that, since I can reproduce the issue!

andermi:
fails on [timing issue here](https://github.com/osrf/buoy_sim/runs/7511716288?check_suite_focus=true#step:4:17190)
I'm not sure why the cpp TestFixture sometimes returns before all iterations have completed.

andermi:
Whew, finally passing 😅

PR #74, opened 2022-07-19
Buoy on RViz Bug fix

quarkytale:
`robot_state_publisher` and `ros_ign_bridge` should not launch by default.

### DIFF
--- a/buoy_gazebo/launch/mbari_wec.launch.py
+++ b/buoy_gazebo/launch/mbari_wec.launch.py
@@ -84,6 +84,7 @@ def generate_launch_description():
             (link_pose_gz_topic, '/tf'),
             (link_pose_gz_topic + '_static', '/tf_static'),
         ],
+        condition=IfCondition(LaunchConfiguration('rviz')),
         parameters=[{'qos_overrides./tf_static.publisher.durability': 'transient_local'}],
         output='screen'
     )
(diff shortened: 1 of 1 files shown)


buoy_gazebo/launch/mbari_wec.launch.py
```diff
@@ -84,6 +84,7 @@ def generate_launch_description():
             (link_pose_gz_topic, '/tf'),
             (link_pose_gz_topic + '_static', '/tf_static'),
         ],
+        condition=IfCondition(LaunchConfiguration('rviz')),
```
chapulina, line 87: Just a nitpick that the TF and robot description stuff is useful outside of RViz. People wanting to use ROS controllers and other ROS tooling will most likely need it, even if they're not visualizing through RViz.

Do you think that exposing a separate parameter, like `enable_ros` or something like that, would make sense here? Even though I wonder why someone would use ROS launch but not want to expose this functionality. They can always just call `ign gazebo mbari_wec.sdf` if they want just the simulator 🤔

quarkytale: That was the intention to not condition it earlier. This was to fix the issue Michael was facing of robot_state_publisher throwing errors, `[robot_state_publisher-3] Error:   Could not find the 'robot' element in the xml file` but I couldn't replicate it. @andermi are you still blocked by that?

But yeah you're right, I guess we could keep this out for now and revisit later, I'll keep an eye if that resurfaces.

andermi: I'm not seeing the `robot_state_publisher` issue any more

quarkytale: I'll close this for now then

PR #78, opened 2022-08-05
Reorganize plugins into different models

andermi:
reorganized models with plugins

- `mbari_wec_base`: raw model with links and joints
- `mbari_wec`: add multiphysics plugins for air spring and EH-PTO
- `mbari_wec_ros`: add ros2 controllers

inheritance:
`mbari_wec_base` -> `mbari_wec` -> `mbari_wec_ros`

`worlds/mbari_wec.sdf` uses `mbari_wec_ros` and adds the joint_state stuff


buoy_description/models/mbari_wec_ros/model.config
```diff
@@ -0,0 +1,24 @@
...
+    <email>louise@openrobotics.org</email>
+  </author>
+  <author>
+    <name>Dharini Dutia</name>
+    <email>dharini@openrobotics.org</email>
+  </author>
+
+  <description>
+    ROS2-enabled MBARI-WEC (Wave Energy Converter) model for ocean sensing and harvesting wave energy
+    with quasi-static multi-physics plugins for Pneumatic Spring and Electro-Hydraulic Power Take-Off.
```
chapulina, line 22: I think it's worth mentioning that the plugins necessary to get TF, `robot_state_publisher` and odometry from ROS aren't included.

chapulina (commented):
I haven't tried running it but the idea LGTM 👍🏽

buoy_description/models/mbari_wec_ros/model.config
andermi: added
- `JointStatePublisher`
- `PosePublisher`
- `OdometryPublisher`

quarkytale (approved):
Works for me

PR #80, opened 2022-08-09
Updating worlds after model reorg

quarkytale:
- Bug fix for robot_state_publisher dying, unable to find the model after reorganization.
 ```
Error [SDF.cc:160] Tried to use callback in sdf::findFile(), but the callback is empty.  Did you call sdf::setFindCallback()?
Error Code 13: [/sdf/world[@name="world_demo"]/model[@name="MBARI_WEC_ROS"]/include[0]/uri:/home/anderson/wec/igngzb/openrobotics/workspace/src/buoy_sim/buoy_gazebo/worlds/mbari_wec.sdf:L49]: Msg: Unable to find 
```
- Updating other worlds, launch files and making sure they work as expected.

### DIFF
--- a/buoy_gazebo/CMakeLists.txt
+++ b/buoy_gazebo/CMakeLists.txt
@@ -15,8 +15,9 @@ else()
 endif()
 
 find_package(ament_cmake REQUIRED)
-find_package(rclcpp REQUIRED)
 find_package(buoy_msgs REQUIRED)
+find_package(buoy_description REQUIRED)
+find_package(rclcpp REQUIRED)
 
 find_package(ignition-cmake2 REQUIRED)
 find_package(ignition-plugin1 REQUIRED COMPONENTS register)
--- a/buoy_gazebo/CMakeLists.txt
+++ b/buoy_gazebo/CMakeLists.txt
@@ -15,8 +15,8 @@ else()
 endif()
 
 find_package(ament_cmake REQUIRED)
-find_package(buoy_msgs REQUIRED)
 find_package(buoy_description REQUIRED)
+find_package(buoy_msgs REQUIRED)
 find_package(rclcpp REQUIRED)
 
 find_package(ignition-cmake2 REQUIRED)
--- a/buoy_description/CMakeLists.txt
+++ b/buoy_description/CMakeLists.txt
@@ -3,14 +3,18 @@ project(buoy_description)
 find_package(ament_cmake REQUIRED)
 
 # Model Generation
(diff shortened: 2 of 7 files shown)

### CODE REVIEW

**Comment by quarkytale** [comment]
So rviz isn't running as expected robot_state_publisher is looking for mbari_wec model as opposed to mbari_wec_ros, sorry this error slipped from me earlier. I'm still debugging but opening this up if anyone can take a glance.

File: buoy_gazebo/CMakeLists.txt [hunk 1]
```diff
@@ -15,8 +15,9 @@ else()
 endif()
 
 find_package(ament_cmake REQUIRED)
-find_package(rclcpp REQUIRED)
 find_package(buoy_msgs REQUIRED)
+find_package(buoy_description REQUIRED)
```
chapulina, line 19: We could move this up for alphabetic order

andermi:
Unfortunately, the error is back that I mentioned in #74 , and I can't figure out what's causing it

```
Error:   Could not find the 'robot' element in the xml file
         at line 80 in /tmp/binarydeb/ros-galactic-urdfdom-2.3.5/urdf_parser/src/model.cpp
Failed to parse robot description using: urdf_xml_parser/URDFXMLParser
terminate called after throwing an instance of 'std::runtime_error'
  what():  Unable to initialize urdf::model from robot description
```

quarkytale:
> I actually can't find where the `parser plugin` is actually converting `sdf` to `urdf` referenced by this comment:

When a SDF if published on the `robot_description` topic, `robot_state_publisher` checks in the list of available parsers which can parse the message and uses `sdformat_urdf` from that list (afaik).


This is actually a very inconsistent error. Thank you for opening a bug report, I'll take that forward!

quarkytale:
Some updates on this
- The issue with `ign sdf` callback failing is SDF can't parse the `package://` path. A quick fix is to create and set an `env` variable `SDF_PATH` the same as `IGN_GAZEBO_RESOURCE_PATH`. This seems like a bug I'll take it up.
- With the `Could not find the 'robot' element in the xml file` issue, its essentially the workspace not being sourced or `sdformat_urdf` not installed. Make sure to install deps via `rosdep` too.

quarkytale:
Just pushed a fix, needed to add another path to the `IGN_GAZEBO_RESOURCE_PATH`, solves the rviz problem as well.

andermi:
The tests should pass after I merge https://github.com/osrf/buoy_sim/pull/85 and you merge main into your branch

PR #87, opened 2022-08-24
ElectroHydraulicPTO solver stability

andermi:
Something was causing numerical instability in the ElectroHydraulicPTO "rpm/pressure -> winding current -> torque" solution

To start, I updated the volumetric and mechanical efficiency maps per vendor efficiency maps codified in matlab by @hamilton8415 and brought in a different interpolator: [Splinter](https://github.com/bgrimstad/splinter/blob/master/docs/cpp_interface.md)

@hamilton8415 and I will investigate/resolve issues in this PR


andermi:
@hamilton8415 something else that was crashing... when the discriminant was negative in the solution for `VBus`

https://github.com/osrf/buoy_sim/blob/af87b95ea6e4e1a3e67385985cec6d9f1adfb8ed/buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicPTO.cpp#L335-L337

andermi:
requires https://github.com/osrf/buoy_msgs/pull/26

andermi:
@hamilton8415 can you check what I did with the new interp and see if the plugin produces the correct output?

hamilton8415:
Will do, I actually wanted to speak with you about how best to incorporate the test comparisons I previously did with the data from the test machine.

andermi:
@mjcarroll @chapulina @quarkytale the new interpolator (if we decide to keep it) has an [MPL-2.0 License](https://github.com/bgrimstad/splinter/blob/master/LICENSE), and I was wondering the best way to credit their project?

chapulina:
IANAL, but reading [Q9](https://www.mozilla.org/en-US/MPL/2.0/FAQ/) here, I think all we need is to include the license file? @tfoote may have a better idea.

tfoote:
Yes, we should bring in the license copy too.

For clarity I would actually suggest packing splinter as a separate package and then just using it in this package to allow easier tracking of potential upstream changes as well as keeping the licensing differentiation clearer.

andermi:
> splinter as a separate package

I've got it in its own package now.
TODO: I assume I should set the license of the package to MPL-2.0 and put the license file at the top level of the package

andermi:
requires https://github.com/osrf/buoy_msgs/pull/28

buoy_gazebo/launch/mbari_wec.launch.py
```diff
@@ -60,7 +65,8 @@ def generate_launch_description():
             pkg_buoy_gazebo,
             'worlds',
             LaunchConfiguration('world_file')
-        ])}.items(),
+        ]),
+            'debugger': LaunchConfiguration('debugger')}.items(),
```
mjcarroll, line 69:
```suggestion
        ]),
        'debugger': LaunchConfiguration('debugger')}.items(),
```

buoy_gazebo/launch/mbari_wec_sinusoidal.launch.py
```diff
@@ -60,7 +65,8 @@ def generate_launch_description():
             pkg_buoy_gazebo,
             'worlds',
             LaunchConfiguration('world_file')
-        ])}.items(),
+        ]),
+            'debugger': LaunchConfiguration('debugger')}.items(),
```
mjcarroll, line 69:
```suggestion
        ]),
        'debugger': LaunchConfiguration('debugger')}.items(),
```

buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicSoln.hpp
```diff
@@ -167,6 +184,101 @@ const std::vector<std::vector<double>> ElectroHydraulicSoln::Neff
...
+    0.9580, 0.9580, 0.9580, 0.9580, 0.9580, 0.9580, 0.9580, 0.9580, 0.9580, 0.9580},
+  {0.1000, 0.7500, 0.8540, 0.8920, 0.9120, 0.9260, 0.9360, 0.9460, 0.9460, 0.9560, 0.9560, 0.9560,
+    0.9560, 0.9560, 0.9560, 0.9560, 0.9560, 0.9560, 0.9560, 0.9560, 0.9560, 0.9560},
+  {0.1000, 0.7400, 0.8460, 0.8880, 0.9080, 0.9240, 0.9340, 0.9440, 0.9440, 0.9540, 0.9540, 0.9540,
+    0.9540, 0.9540, 0.9540, 0.9540, 0.9540, 0.9540, 0.9540, 0.9540, 0.9540, 0.9540},
+  {0.1000, 0.7300, 0.8380, 0.8840, 0.9040, 0.9220, 0.9320, 0.9420, 0.9420, 0.9520, 0.9520, 0.9520,
+    0.9520, 0.9520, 0.9520, 0.9520, 0.9520, 0.9520, 0.9520, 0.9520, 0.9520, 0.9520},
+  {0.1000, 0.7200, 0.8300, 0.8800, 0.9000, 0.9200, 0.9300, 0.9400, 0.9400, 0.9500, 0.9500, 0.9500,
+    0.9500, 0.9500, 0.9500, 0.9500, 0.9500, 0.9500, 0.9500, 0.9500, 0.9500, 0.9500}
+};
```
mjcarroll, line 233: Does it make sense to put these in a format that can be read off disk or are they not changing frequently?

buoy_tests/tests/pc_commands_ros_feedback.cpp
```diff
@@ -201,6 +201,8 @@ TEST_F(BuoyPCTests, PCCommandsInROSFeedback)
     static_cast<int>(node->clock_->now().seconds()),
     static_cast<int>(iterations / 1000.0F));
 
+  std::cout << node->torque_policy_ << std::endl;
```
mjcarroll, line 204: Is this spurious debug output?

buoy_gazebo/src/ElectroHydraulicPTO/ElectroHydraulicSoln.hpp
hamilton8415: These values are very specific to the hydraulic motor used, so from my point of view it's OK to hard-code them, they should never change for modelling this equipment, if the plugin became more general purpose then maybe, but at this point there's a lot of hardware-specific items in defines and such...

buoy_tests/tests/pc_commands_ros_feedback.cpp
andermi: I can take it out, I just liked seeing in the logs what settings were used

andermi:
I'm gonna say we should merge this in soon as this has been out quite a while and more branches are starting to use it as a base.

buoy_tests/tests/pc_commands_ros_feedback.cpp
andermi: I think I might just leave it in to log policy settings used in the test

andermi: actually, just discovered the config from param wasn't being loaded... fixing now

mjcarroll:
> We should merge this in soon as this has been out quite a while and more branches are starting to use it as a base.

Agree, what other things need to get in to call this done?

andermi:
> what other things

I think we are good as is. @hamilton8415 has other branches with some other tweaks coming soon

quarkytale:
Works for me, CI should come out green as well

mjcarroll (approved):

andermi:
@hamilton8415 has requested I cherry-pick new motor efficiency values into this PR before we merge.

PR #88, opened 2022-09-01
SpringController in mins

quarkytale:
Support for `buoy_msgs` [#26](https://github.com/osrf/buoy_msgs/pull/26) and [#27](https://github.com/osrf/buoy_msgs/pull/27). Pulled these out from the solver stability PR since that seems like will take a while.


andermi:
I actually made these changes and a few other related fixes (like in the tests) in #87 but we could pull those out since that PR might be open for a bit longer.

quarkytale:
Yeah that's what I was aiming for. Some tests are failing, retriggered CI.

Issue #43, opened 2022-05-14
Jitter in ForceTorque & SCRecord upper/lower_psi


andermi:
I have added the ForceTorque sensor and plugin and am subscribing to it in the SpringController and publishing out to ros2 /sc_record!

![load_cell](https://user-images.githubusercontent.com/20071132/167954173-7d39dc8b-8072-4bd9-8610-aeb55e5b0248.png)
One issue though, the plot shows that the force jumps between two values (quantization or resolution issue?) at steady-state. The jump at steady-state is quite large: from 16700 to 17400.

_Originally posted by @andermi in https://github.com/osrf/buoy_sim/pull/39#discussion_r870785861_

        @chapulina I wonder if it's related to this as well: I noticed the same thing happens in PolytropicPneumaticSpring when I plot the pressure of the chamber. The plot shows a jitter of 50kPa.

![Screenshot from 2022-05-13 12-38-57](https://user-images.githubusercontent.com/20071132/168378953-ea659438-6f7c-4924-9d44-8d3af4471452.png)

Pressure is computed with the piston joint position:
`_ecm.Component<ignition::gazebo::components::JointPosition>(this->dataPtr->jointEntity)`

_Originally posted by @andermi in https://github.com/osrf/buoy_sim/pull/39#discussion_r872740989_

andermi:
from @chapulina in #39:

> > the force jumps between two values (quantization or resolution issue?) at steady-state.
> 
> I tracked this down to DART's [getBodyForce](https://github.com/dartsim/dart/blob/876567c7f2e65b6678c7249e4f9a18271bc1863d/dart/dynamics/BodyNode.hpp#L784-L789) call, and also checked that it happens with both solvers: `dantzig` (default) and `pgs` (used in #26 ).
> 
> We can ticket an issue to dig deeper into this if you think it will cause problems here.

PR #94, opened 2022-09-28
Sim Spring Comparison with Test Machine

andermi:
compare sim spring with test machine data

- [x] fix polytropic curves
- [x] init piston position to align with test data
- [x] use common google test fixture

to run the manual mode:
`ros2 launch buoy_tests experiment_comparison_select.launch.py manual:=true`
or comparison mode:
`ros2 launch buoy_tests experiment_comparison_select.launch.py manual:=false`
or run the launch test (which CI will pick up):
`launch_test install/buoy_tests/share/buoy_tests/launch/experiment_comparison.launch.py`
which defaults to running select with `manual:=false`


hamilton8415:
Very good, cool to see the pressures matching decently.  I see that you created a second test for the spring components, I am not sure of the pro/cons of combining this with the electro-hydraulic tests, it could be done and since a lot of the data reading, plotting, writing is the same in each.  My guess is you've thought about this, I'm flexible either way.  As separate tests though, we need to distinguish the output files from pto_experiment_comparison and spring_experiment_comparison.  At the moment they are using the same names, which means the output of one test is likely to be read by the other, which doesn't make sense.  That would be an argument for combining and not having to address this.  OTOH, it'll be easier to track down issues if they are separate...    Also, I was interested to see you writing the upper and lower chamber volumes out as new fields, neccessitating a different number of fields to be read in in the .exp case versus the .tst case.  I'm sure there's a reason, but it seems these volumes could be re-computed in the .tst case the same as the .exp case.  Finally, the position plots are off by a couple of inches, which translates to the results.  The experiment data definately starts at 2.9 inches or so as a result of how the test machine works, and you set that offset, but it doesn't seem to hold, looking at the .tst output the first line has the offset, and then it jumps back to zero.   Looks like  you set the initial position and then let gazebo integrate the specified velocity forward to provide the position.  That seems like it should work, but not sure why the position reverts after the first step.  Is it possible to specify the position at each timestep instead of the velocity.  Would preclude combining the tests somewhat of course so perhaps that's what you were thinking...

hamilton8415:
Also curious about the noise in the pressure signals at rest around 150psi in the lower spring pressure, surprising to see that in the simulated data.  But, those issues aren't really related to this PR as I believe we've been telling ourselves we'll tackle inaccuracies in the simulation subsequent to getting these tests working.  Ha, hard to resist though once you see the graph comparisons...

andermi:
> cool to see the pressures matching decently
They are close, but I used some other data I happened to find to fit the polytropic index and the set points for P, V so I'm not surprised its a bit off. I can refit the new polytropic curves to the test bench easily enough.

Upper_PV
Lower_PV

> created a second test for the spring components
Yes, I did that for now (save some effort to see how things are laid out), but I think we should make a [Test Fixture](https://github.com/google/googletest/blob/main/docs/primer.md#test-fixtures-using-the-same-data-configuration-for-multiple-tests-same-data-multiple-tests) to share the data reading code and other utility funcs between the two. We would then have two separate `TEST_F` blocks to handle differences.

> using the same names
I'll fix that.

> volumes could be re-computed in the .tst case the same as the .exp case
good idea

> experiment data definitely starts at 2.9 inches
fixed the starting point and they match much better -- I was setting to zero

> curious about the noise in the pressure signals
This has been present in the data since the beginning and I noted the issue here: https://github.com/osrf/buoy_sim/issues/43
and seems to be coming from something in DART

andermi:
image
Screenshot from 2022-09-28 15-10-31
image
Screenshot from 2022-09-28 15-12-15

fixed polytropic curves

andermi:
right now, the jitter in pressure is not consistent between two runs and causes the comparison test to fail

andermi:
when @hamilton8415 fixes linter errors in `pto_experiment_comparison.cpp`, I'll merge his changes and non test-related errors should go away

andermi:
> fixes linter errors in `pto_experiment_comparison.cpp`

@hamilton8415 although, it might be wasted time if I merge our tests

hamilton8415:
I'm flexible.  I think if we can save some work by merging the tests with regard to setting up a shared test-fixture, etc.  then that's what we should do unless we can think of a specific reason it's a bad idea.  In the meantime, I'll take a look at the linter errors as I need a better understanding of that anyway.

hamilton8415:
Very good, I ran this, tests pass and it all looks great, nice to clean up the goof-ball way I had been switching between manual and auto tests...  Seems like it's time to merge this into pto_experiment_comparison, I will do that and in the coming week we can then merge that into main after removing the individual pto and spring test source files.  Thanks andermi for the excellent work.

hamilton8415:
Actually, I will wait a bit into this week to give quarkytale and mjcarroll a chance to review, I missed that request from andermi.  Have a great weekend!

andermi:
> time to merge this into pto_experiment_comparison

Given the type of changes here, it may be easier to merge the other way into this branch and continue your work off of this branch. Then merge this branch directly to main.

buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
```diff
@@ -97,9 +100,74 @@ class WindingCurrentTarget
...
+//                   \                      |                                    \
+//                    \                     |                                     \
+//            -35A ->  ---------------------|------------------------------------------
+//                     ^                    |                                      ^
+//                     |                    |                                      |
+//                 -5000RPM                 V                                   6000RPM
+
+    double AdjustedN = N;
+    if (fabs(N) >= 0) {  // Retracting
+      if (RamPosition < (STOP_RANGE - SC_RANGE_MIN)) {
```
andermi, line 137: @hamilton8415 it doesn't look like `RamPosition` is being updated by anything. Should this be set by `ignition::gazebo::components::JointPosition` in `ElectroHydraulicPTO.cpp`?

hamilton8415 (approved):
Great!

buoy_gazebo/src/ElectroHydraulicPTO/WindingCurrentTarget.hpp
andermi: @hamilton8415 hooking up the `RamPosition` to the position data from the `ECM` introduced some problems with the winding current limiter and RPM so, I'll leave that for you to change in the next branch with a `TODO(hamilton)`

hamilton8415 (approved):
Looks like maybe you found why the mean spring position was different.  Still approve... :)

hamilton8415:
Very Nice! I'lll run this and check it out. Have a great weekend. -- ah 


_______________________________________________ 
Andrew Hamilton, Ph.D. 
Engineering Division Chair 
Monterey Bay Aquarium Research Institute 
7700 Sandholdt Road, Moss Landing, CA 95039 
Phone: 831-775-1861 Fax: 831-775-1646 
http://www.mbari.org/

PR #100, opened 2022-10-11
Adding Hydraulic Pneumatic Friction plugin

quarkytale:
A new plugin and test framework is ready. Need to work on model accuracy.


hamilton8415:
Nice job Dharini making a clean and simple plugin to address this need.  I have a few suggestions for checking/consideration.   The friction force versus speed interpolation defined by:

 pistonSpeed{-0.4F, -0.3F, -0.2F, -0.1F, 0.0F, 0.1F, 0.2F, 0.3F, 0.4F},
 meanFriction{1200.0F, 1000.0F, 700.0F, 500.0F, 0.0F, -1000.0F, -1400.0F, -2100.0F, -2900.0F} 

probably needs a larger extend, in the paper tests were done up to these speeds (+/- 0.4 m/s) but in actuality the piston could go quite a bit faster.  I'd say to be sure of covering the possibilities, the chart should go up to +/- 2m/s.  Because this interpolation consists of just a few straight lines, it may not be necessary to include points in the intermediate parts of the straight sections, which could speed up the interpolation some.  I  believe splinter may be configured to use the highest values in the interpolation table when independent variables greater than the defined extent are used, this would be good to include if it's possible.  @andermi knows what's possible I believe.

Also, in what's sure to be a continual source of confusion, the internal gazebo model uses and opposite sign for the position of the ram extension compared with the paper and actual system at sea.  So we need to swap the sign of the velocity relative to the paper.  For now I suggest we keep this annoyance as we've addressed it throughout the code in a number of places, if we decide to change it to match, we can do it once and hopefully find all places to fix it.

Finally, I will suggest a simpler name for this plugin.  I think PTO_Friction or similar is fine, the friction being simulated here is a combination of a lot of contributions, from the hydraulic ram, from the pneumatic part, from extraneous other seals, etc.  So avoiding the specificity of HydraulicPneumatic might be a good idea.  Not too important if it's challenging to change.

quarkytale:
That's exactly the feedback I was looking for, thanks Andy! 
I saw the behaviour with speed outside the bounds I modelled for was kinda weird. I would want to add all the values inclusive of the range speed will touch. Love some help on getting that data. Otherwise, I'm not sure how to configure the splinter to work that case, will ping Michael!
Changes upcoming:
- Switching the speed's sign
- Changing plugin name to PTOFriction

hamilton8415:
Michael is out on the boat today, so here's an example of splinter use from elsewhere that returns the endpoints when an out of bound request is made:
 const double eff_m = this->hyd_eff_m.eval(rpm, splinter_ros::USE_BOUNDS);  

I haven't tested this carefully, (in michael we trust...:), but in the friction case we'd still want a wide expanse and only count on this for the unexpected case of a higher speed than expected.  Otherwise I'm not sure what spinter returns when the argument is out of bounds, perhaps it extrapolates, or returns dragons...

quarkytale:
Update the plugin as per feedback. 
The model is still very inaccurate:
```
/home/osrf/buoy_ws/src/buoy_sim/buoy_gazebo/test/test_pto_friction.cpp:41: Failure
2: The difference between 0.0 and friction_model.eval(0.0, splinter_ros::USE_BOUNDS) is 150.00000000000006, which exceeds 0.05, where
2: 0.0 evaluates to 0,
2: friction_model.eval(0.0, splinter_ros::USE_BOUNDS) evaluates to -150.00000000000006, and
2: 0.05 evaluates to 0.050000000000000003.
2: /home/osrf/buoy_ws/src/buoy_sim/buoy_gazebo/test/test_pto_friction.cpp:42: Failure
2: The difference between -2100.0 and friction_model.eval(-0.3, splinter_ros::USE_BOUNDS) is 193.74997289851353, which exceeds 100.0, where
2: -2100.0 evaluates to -2100,
2: friction_model.eval(-0.3, splinter_ros::USE_BOUNDS) evaluates to -2293.7499728985135, and
2: 100.0 evaluates to 100.
```
What's our tolerance?

buoy_gazebo/src/PTOFriction/PTOFriction.cpp
```diff
@@ -0,0 +1,158 @@
...
+  const std::vector<double> meanFriction;  // N
+
+  /// \brief Construct and approximation of friction model using linear spline
+  splinter_ros::Splinter1d pto_friction_model;
+
+  const splinter_ros::FillMode & fill_mode = splinter_ros::USE_BOUNDS;
+
+  PTOFrictionPrivate()
+  : pistonSpeed{0.4F, 0.1F, -0.1F, -0.4F},
+    meanFriction{1200.0F, 700.0F, -1000.0F, -2900.0F},
```
andermi, line 58: if the vector is type `double`, omit the `F` to initialize with `double` literals rather than `float` literals

andermi:
> The model is still very inaccurate:

I'm going to open a PR in a moment to add a different interpolator. We have found issues with splinter not doing the right thing, unfortunately.

buoy_gazebo/src/PTOFriction/PTOFriction.cpp
andermi: Also, I think the sign of the piston speed is flipped:

```
  : pistonSpeed{-0.4F, -0.1F, 0.1F, 0.4F},
    meanFriction{1200.0F, 700.0F, -1000.0F, -2900.0F},
```

andermi: it looks like a couple points could be added:

(-0.05, 400)  // endpoint of the left line
(0.0, 0.0)  // we'd like the applied force to be zero when the piston isn't in motion
(0.05, -750)  // starting point of the right line

buoy_gazebo/src/PTOFriction/PTOFriction.cpp
```diff
@@ -0,0 +1,158 @@
...
+  // We just created the joint velocity component, give one iteration for the
+  // physics system to update its size
+  if (prismaticJointVelComp == nullptr || prismaticJointVelComp->Data().empty()) {
+    return;
+  }
+
+  // Interpolate the new friction force based on current joint velocity
+  auto friction_force =
+    this->dataPtr->pto_friction_model.eval(
+    fabs(prismaticJointVelComp->Data().at(0)), this->dataPtr->fill_mode);
```
andermi, line 138: Since you have both negative speed and positive speed in your interpolation table (since they are different lines), don't take `fabs` here. Also, instead of flipping the sign in the table to account for the `x` direction difference between the sim and physical buoy, negate the velocity here.

```
-prismaticJointVelComp->Data().at(0)
```

buoy_gazebo/src/PTOFriction/PTOFriction.cpp
andermi: also, extrapolating out to +/- 5 m/s as Andy suggests, I compute the following points:

(-5.0, 12750)
(5.0, 32033)

mjcarroll:
Pending removing splinter.

mjcarroll:
https://github.com/osrf/buoy_msgs/pull/36 has been merged, so you should be able to update accordingly here.

andermi (approved):
LGTM once splinter is replaced :+1:

quarkytale:
Replaced splinter, but the build will totally fail without the motor mode PR.

andermi:
> build will totally fail without the motor mode PR

Sorry, I didn't quite think that one through. The issue is that `ElectroHydraulicPTO` plugin still needs `splinter_ros`? ... and we can't remove the `splinter_ros` dependency without the switch to 1d interpolation on mechanical and volumetric efficiencies from motormode

hamilton8415:
Agreed, I am running down a couple of test issues with motormode but let's aim to get that merged this week so we can have a buildable main again.

quarkytale:
buoy sc tests are failing since buoy position is now different with the friction added, working on updating the numbers.

buoy_gazebo/CMakeLists.txt
```diff
@@ -141,6 +141,60 @@ install(DIRECTORY
...
+      add_launch_test(test/launch/${TEST_NAME}.launch.py
+        TIMEOUT 300
+      )
+    else()
+      ament_add_gtest_test(${TEST_NAME})
+    endif()
+  endfunction()
+
+  # Add gtests
+  gz_add_gtest(test_pto_friction ROS)
```
andermi, line 196: Please, move the test from `buoy_gazebo` to `buoy_tests`.

buoy_gazebo/CMakeLists.txt
```diff
@@ -141,6 +141,60 @@ install(DIRECTORY
...
+    ament_add_gtest_executable(${TEST_NAME}
+      test/${TEST_NAME}.cpp
+    )
+    if(gz_add_test_GAZEBO)
+      target_link_libraries(${TEST_NAME}
+        ignition-gazebo${IGN_GAZEBO_VER}::ignition-gazebo${IGN_GAZEBO_VER}
+      )
+    endif()
+    if(gz_add_test_ROS)
+      set(ROS_PKGS rclcpp simple_interp ${gz_add_test_EXTRA_ROS_PKGS})
```
andermi, line 175: `simple_interp` should be added as an argument to the `gz_add_gtest` function after `EXTRA_ROS_PKGS` instead of explicitly adding it here.

buoy_gazebo/CMakeLists.txt
andermi, line 196: use all of this in `buoy_tests` instead

buoy_gazebo/CMakeLists.txt
andermi: In order to use your include files for the friction plugin in the test in `buoy_tests`, copy the pattern [here](https://github.com/osrf/buoy_sim/blob/55dcc0c0ce1728034a49cdacb3d711ab9dbdda2a/buoy_gazebo/CMakeLists.txt#L125-L128) to install your include files

buoy_gazebo/test/test_pto_friction.cpp
```diff
@@ -0,0 +1,47 @@
...
+  const std::vector<double> force;  // N
+  simple_interp::Interp1d friction_model;
+
+  PTOFrictionTest()
+  :  speed{-5.0, -0.4, -0.1, -0.05, 0.0, 0.05, 0.1, 5.0},
+    force{12750.0, 1200.0, 700.0, 400.0, 0.0, -750.0, -1000.0, -32033.0},
+    friction_model(speed, force)
+  {
+  }
+};
```
andermi, line 37: Try including your plugin class rather than recreate here so we test the values in used in the plugin.

buoy_gazebo/test/test_pto_friction.cpp
```diff
@@ -0,0 +1,47 @@
+// Copyright 2020 Open Source Robotics Foundation, Inc.
```
andermi, line 1: move this whole test file to `buoy_tests`

buoy_gazebo/test/test_pto_friction.cpp
```diff
@@ -0,0 +1,47 @@
...
+
+  PTOFrictionTest()
+  :  speed{-5.0, -0.4, -0.1, -0.05, 0.0, 0.05, 0.1, 5.0},
+    force{12750.0, 1200.0, 700.0, 400.0, 0.0, -750.0, -1000.0, -32033.0},
+    friction_model(speed, force)
+  {
+  }
+};
+
+TEST_F(PTOFrictionTest, ModelAccuracy)
```
andermi, line 39: This is a good test for the interpolator.
I wonder if it would be good to add a test like:

1. run the sim without the friction plugin disabled and measure settling time
2. run the sim with the friction plugin enabled and the settling time should be faster

andermi (changes requested):
I just caught a couple more things

buoy_gazebo/CMakeLists.txt
andermi: Sorry, I was wrong about copying that pattern to use the includes. (I wasn't thinking quite clearly!)

You need to export the friction plugin library

andermi: but that's only if you want to use the class directly to recreate the test you made...
Or, you could just use the gz_sim test fixture and load the plugin that way for other testing

quarkytale:
Changing absement values is not affecting the piston motion

andermi:
> Changing absement values is not affecting the piston motion

fixed with 1ebd6c6

andermi (approved):
LGTM

PR #101, opened 2022-10-19
Custom Interp1d

andermi:
Unfortunately, splinter is just plain wrong in 1D and 2D. We reformulated the efficiency tables in the EHPTO to be 1D, so, now all that's needed is simple linear 1D interpolation.

I kept the same interface as before, but this time `USE_BOUNDS` is always enabled and those extra arguments for `fill_mode` are removed


hamilton8415 (approved):
Thanks for doing this, nice to have an interpolator that works and we have some visibility into.  For the foreseeable future I think we can try to stay with only needing 1-d interpolation, higher dimensions are always going to be expensive.  I did use this in the motormode branch and had no problems.  One could consider overloading () for the evaluation, that is neat and is how the Justinterp library did it, although I'm not sure if it makes the use more or less clear.  I'm flexible on this point for sure.

andermi:
> One could consider overloading () for the evaluation

added `operator()` for @hamilton8415 ;)

PR #102, opened 2022-10-28
Refine Spring

andermi:
The spring model was causing some jitter in the piston position, spring pressures, as well as the force-torque sensor. This was due to a discontinuity in pressures in the implementation of hysteresis in the spring. This agitation in forces occurred for very small velocities (based on changes of sign). A dead-band was added to checking velocity to fix this. Also, pressure now evolves continuously across the hysteresis rather than a step change.

Additionally, to better model the thermal decay of pressure at low speeds, a cooling law has been implemented within the velocity dead-band.

image
image
image
image

### DIFF
--- a/buoy_description/models/mbari_wec/model.sdf
+++ b/buoy_description/models/mbari_wec/model.sdf
@@ -19,8 +19,8 @@
       <chamber>upper_polytropic</chamber>
       <is_upper>true</is_upper>
       <!-- measure of valve opening cross-section and duration (meter-seconds) -->
-      <valve_absement>49e-7</valve_absement>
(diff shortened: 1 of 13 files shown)


hamilton8415 (approved):
You're showing off a little with those sick graphs, nice job!   Shortly here I'll provide some test data from an at-sea deployment, it may be worth tuning to that at some point as the larger heat-sink of the ocean could be relevant.  Also the usual behaviors at sea have a lower frequency content than the friction moves in the lab.  Approved in any event.

andermi:
Hah! plotjuggler deserves most of the credit ;-p

PR #83, opened 2022-08-17
Upgrade to Humble + Garden

chapulina:
* Part of https://github.com/osrf/buoy_sim/issues/81

WIP

PR #103, opened 2022-11-03
Hamilton8415/fs hydrodynamics

hamilton8415:
Pull request for free-surface hydrodynamics.  Still WIP.  Currently this plugin implements the linearized wave-radiation forces using WAMIT data that is read in from a file, and linearized buoyancy.  The wave-exciting forces are also read from WAMIT but are not currently included as there's some work and testing to do there.  Plenty of review needed, particularly in the manner that the WAMIT data is stored and read, among other items.


srmainwaring:
@hamilton8415 and @mjcarroll, as an FYI I've created a version of this branch rebased on #83 here: https://github.com/srmainwaring/buoy_sim/tree/srmainwaring/humble-garden-fs_hydro-2212. It's required for building on macOS which needs humble-garden, but may be of use when #83 is merged if you decide to then rebase to clean up the history rather than merge. The comment in #83 about the gnuplot-iostream dependency applies here too.

hamilton8415:
With the added-mass capabilities now in the garden binaries, this PR is now able to be merged.  Thanks go to @andermi for fixing a few issues recently.  A number of tasks remain so I'm listing them here:

- [ ] Add ability to ingest a seed value from the .sdf value, and use that to set the irregular waves.  This will require an update to the FreeSurfaceHydrodynamics library to implement, but should be straightforward. https://github.com/hamilton8415/FreeSurfaceHydrodynamics
- [ ] Pass existing tests.  With this plug-in in the system, the buoy will start moving as soon as the sim starts, so any tests that relied on the buoy falling from a small height and then coming to rest will not work correctly.  If the incident wave-heights are set to zero, the general behavior we're used to of the buoy oscillating a bit and coming to rest should still be present.  The rates of this will be changed as there is now wave-radiation damping added to the system...  I am not entirely sure how to implement this in the CI tests.
- [ ] This plugin needs some tests, currently the FreeSurfaceHydrodynamics has some non-CI based tests that can be helpful here.  In particular that library can compute the time-domain motion of a floating body starting from rest in a prescribed sea-state,  for a single body, but independently of gazebo/DART.  So a great test will be to make sure the gazebo simulation results for a single floating body match those results decently.  Should be nearly perfect if all the physical characteristics of the body are the same, the only difference would be different time-step size or ODE solver technique between the two softwares.  This test will also be a good additional check on the added mass features in Gazebo.
- [ ] This plugin currently blasts out a ton of debug data using a basic std::cout, this needs to be put into better debug messages, etc.
- [ ] There is some messiness in this code related to wave-angle and the buoy yaw.  This is probably OK for the current purpose, but is likely fragile.  As a note, this work has been reduced to being applicable to axi-symmetric bodies, again that's fine for our purpose, and the pieces are here to make it for non-axisymmetric bodies, requiring changes in this plugin and the FreeSurfaceHyrdodynamics library.

andermi:
Oh shoot. I broke CI by using that eta function. Should we build libFreeSurfaceHydrodynamics from source in CI for now anyways? (until you fix debs and release)

hamilton8415:
Yes, thanks for changing the CI to build from source.  That's good, we can adjust the functionality as we need to without changing the debian until we've got it working to our liking.

andermi:
https://user-images.githubusercontent.com/20071132/218591182-da8b9672-25a0-4ffc-92e3-90bed43e1372.mp4

andermi:
@hamilton8415 The error in the `pc_commands_ros_feedback` test:
`[pc_commands_ros_feedback-1] Expected: (node->range_finder_) < (0.97), actual: 1.32002 vs 0.97`
is from buoyancy being disabled in this branch in `server.config`

We can wait for @quarkytale to finish #122 before we worry about that test failure

buoy_tests/testing_utils/utils.py
```diff
@@ -40,9 +47,166 @@
 from rclpy.parameter import Parameter
 
 
+def regenerate_models(context, *args, **kwargs):
```
andermi, line 50: @hamilton8415 I laid the foundation for setting template params in bouy_tests...

buoy_tests/tests/fixture_server.cpp
```diff
@@ -41,7 +40,7 @@ TEST(BuoyTests, RunServer)
 
   // Setup fixture
   gz::sim::ServerConfig config;
-  config.SetSdfFile("mbari_wec.sdf");
+  config.SetSdfFile("mbari_wec_test.sdf");
```
andermi, line 43: @hamilton8415 tests should now refer to `mbari_wec_test.sdf`

buoy_tests/launch/pc_commands_ros_feedback_py.launch.py
```diff
@@ -27,7 +27,13 @@
...
 
 def generate_test_description():
-    return default_generate_test_description()
+    sim_params = dict(inc_wave_spectrum='inc_wave_spectrum_type:None',
+                      physics_rtf=11.0,
+                      physics_step=0.001)
+    return default_generate_test_description(enable_rosbag=True,
+                                             rosbag_name='rosbag2_pc_cmds_py',
+                                             regen_models=True,
+                                             regen_kwargs=sim_params)
```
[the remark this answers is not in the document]
andermi: This test fails in an unexpected way when physics_step=0.01

buoy_tests/testing_utils/utils.py
```diff
@@ -40,9 +47,166 @@
...
+                           bridge,
+                           robot_state_publisher,
+                           rviz]
+        override_params = dict(door_state='open',
+                               inc_wave_spectrum=\
+                                   'inc_wave_spectrum_type:MonoChromatic;A:1.0;T:12.0')
+        launch.actions.OpaqueFunction(function=regenerate_models,
+                                      args=dependent_nodes,
+                                      kwargs=override_params)
+    """
```
andermi, line 65: @hamilton8415 I laid the foundation for setting template params in bouy_tests...

andermi: the python docstring just has a small example of supported overrides... 

I implemented:
- door_state
- physics_step
- physics_rtf
- scale_factor
- inc_wave_seed
- battery_soc
- battery_emf
- inc_wave_spectrum

buoy_tests/tests/wave_body_interactions_heaveonly.cpp
```diff
@@ -0,0 +1,285 @@
...
+    0, 0,
+    2.03);             // Set COG relative to waterplane coordinate system.
+  BuoyA5.SetVolume(1.3658536);
+  BuoyA5.SetMass(buoy_mass);
+
+  //  std::string HydrodynamicsBaseFilename =
+  //      "./example_hydrodynamic_coeffs/BuoyA5";
+  std::string HydrodynamicsBaseFilename =
+    std::string("/home/hamilton/buoy_ws/src/buoy_sim/buoy_description/") +
+    std::string("models/mbari_wec_base/hydrodynamic_coeffs/BuoyA5");
```
andermi, line 223: need to fix hardcoded path for test

buoy_tests/tests/wave_body_interactions_pitchonly.cpp
```diff
@@ -0,0 +1,289 @@
...
+    -.22);             // Set COB relative to waterplane coordinate system.
+  BuoyA5.SetCOG(
+    0, 0,
+    0.0);             // Set COG at waterplane for this test.
+  BuoyA5.SetVolume(2.39);
+  BuoyA5.SetMass(buoy_mass);
+
+  std::string HydrodynamicsBaseFilename =
+    std::string("/home/hamilton/buoy_ws/src/buoy_sim/buoy_description/") +
+    std::string("models/mbari_wec_base/hydrodynamic_coeffs/BuoyA5");
```
andermi, line 223: hardcoded

andermi:
I got things running again with the tests. I still need to update things to make the tests pass now, but at least they run

hamilton8415:
great, thanks for catching the hardcoded path to my machine...

hamilton8415:
I pushed a small change to reduce the verbosity of output from wave_body_interactions_pitch.cpp, and did a little code formatting.  Linters and tests for these still passing.

hamilton8415:
I believe we are finally at the end-zone with this one.  I ran a number of sims today and all worked as expected.  I think we should merge into main and continue testing and comparing to our tuturials from there.  I am going to be away for a bit, so feel free to merge this @andermi as far as I'm concerned.  When this is merged, the analogous PR in buoy_entrypoint should be merged also so that the install instructions remain relevant.

PR #107, opened 2022-11-22
Add mooring

mabelzhang:
Addresses #15 

Opening for visibility. Test isn't complete yet.

Adds mooring line, chain, and seafloor.

2022-11-12-013357_2000x1690_scrot_chain

2022-11-12-015245_2000x1690_scrot_chainJoints

https://user-images.githubusercontent.com/7608908/203235648-7375edbe-d450-42ce-be26-ccc1e6515851.mp4

## Test WIP

The test isn't able to push the buoy all the way to the bounds to test the bounds yet. Because the heave cone is so heavy, a very large force has to be applied to move the buoy just a few meters. After some iterations, an out of bounds error happens, seeming to do with collision, and the pose values go haywire.

3000 iterations of 10,000 N (is that the unit?) in x:
```
[Dbg] [mooring_drift.cpp:106] iter 2275, buoy link pose: 11.9083 0.003313 -2.44625
[Dbg] [mooring_drift.cpp:106] iter 2276, buoy link pose: -2.64042e+39 -9.35589e+38 6.45126e+38
/home/developer/mbari/buoy_sim_ws/src/buoy_sim/buoy_tests/tests/mooring_drift.cpp:108: Failure
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 2.80127e+39 vs 60

ODE INTERNAL ERROR 1: assertion "aabbBound >= dMinIntExact && aabbBound < dMaxIntExact" failed in collide() [collision_space.cpp:460]
Aborted (core dumped)
```

Note how the pose jumps from (12, 0, -2.4) to something huge. That's not the actual test failing (looks for < 60 m from anchor), that's values going out of bounds.

There are a few things I can try:
- Apply ocean currents (https://github.com/gazebosim/gz-sim/pull/800) manually on the command line and see how far that pushes the buoy
- Enable GUI in the test - is there a way to do that? I didn't find a GUI parameter in `ServerConfig` that we can pass to the server.
- Write a standalone plugin (basically identical to the one in the test) to run in the simulation and watch the GUI to check things are moving correctly.

My main uncertainty is that, since I'm not seeing the GUI, even before it goes out of bounds, I don't know if the buoy is moving to the force correctly.


mjcarroll:
* Check performance against main branch
* Check performance with collisions available
* Test instability with ApplyLinkWrench system
* Implement static computation to shortcut dynamics (and as something to check against).

andermi:
> * Implement static computation to shortcut dynamics (and as something to check against).

@mabelzhang Here's a wheel file (in a zip since github doesn't like .whl extension) for a mooring catenary calculator in python I whipped up. We can chat about it at the 2pm, or sooner if you'd like.

unzip and then pip3 install
[mooring_catenary.zip](https://github.com/osrf/buoy_sim/files/10156999/mooring_catenary.zip)

mabelzhang:
I added the first pass of the catenary solver based on Slack discussions. Not sure if I did it correctly.
It's not done yet (hence converted PR to draft). It might need a loop in the outer call to the solver, and the tension and force still need to be calculated from the solver's results.

mabelzhang:
I went back to reread the Python diagram and the code. That made me fix a few things in the C++ solution class that I had misunderstood before.

It doesn't give me NaNs anymore, so I probably fixed something. Without doing anything, B is exactly L - V, which means it's finding a solution at the initial b = 0, hmm maybe that's okay? But c = 0, (since B = L - V, (L-B)^2 - V^2 is exactly 0), which I think is what we don't want? c needs to be > 0 right?

VSolver (I think the fvec[0] here is supposed to be the x=H that is solved):
```
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 5 fvec[0]: 120
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 5 fvec[0]: 120
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 159.431
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 161.174
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 5 fvec[0]: 120
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 159.431
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 161.174
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 157.793
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 165.028
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 152.096
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 189.322
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 137.247
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 4 fvec[0]: 138.85
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 5 fvec[0]: 120
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 5 fvec[0]: 120
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 159.431
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 161.174
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 5 fvec[0]: 120
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 159.431
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 161.174
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 157.793
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 165.028
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 152.096
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 189.322
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 1 fvec[0]: 137.247
[ign gazebo-1] [Dbg] [CatenarySoln.hpp:199] VSolver solverInfo: 4 fvec[0]: 138.85
```
This solverInfo looks okay.

HSolver:
```
[ign gazebo-1] [Dbg] [MooringForce.cpp:183] HSolver solverInfo: 5 c: 0 B: 78
```
This solverInfo is 5, which I think means it didn't find a solution? So maybe I have something else wrong.

Performance is very very slow. Other than I might be doing something wrong, maybe b can be incremented faster.
I can also try moving the buoy to see if I get NaNs, or if B is still reasonable.

andermi:
```

// See MINPACK documentation for detail son this solver
// Parameters and defaults are (Scalar = double):
//           : factor(Scalar(100.))
//           , maxfev(1000)
//           , xtol(std::sqrt(NumTraits<Scalar>::epsilon()))
//           , nb_of_subdiagonals(-1)
//           , nb_of_superdiagonals(-1)
//           , epsfcn(Scalar(0.)) {}
  Eigen::HybridNonLinearSolver<ElectroHydraulicSoln> solver(this->dataPtr->functor);
  solver.parameters.xtol = 0.0001;
  solver.parameters.maxfev = 1000;
  solver.diag.setConstant(3, 1.);
  solver.useExternalScaling = true;       // Improves solution stability dramatically.
```

Try `useExternalScaling=true`

mabelzhang:
With these lines locally
```
    vSolver.diag.setConstant(1, 1.);
    vSolver.useExternalScaling = true;
```
It didn't seem to make a difference to the results or the speed.

I think I must be doing something wrong in the HSolver, for it to always get solverInfo = 5... maybe I need to check my operator() again.

Tell me if this sounds wrong. The inner optimization wants y(H)-V, where y is the cosh function, so it sets fvec[0]=y(H)-V at the end. The outer optimization wants y_B(B)=H, so it first calls the inner class with initial guess x0=H, and once the inner optimization is done, it uses whatever x that is found, and just returns that? Or should it be returning y_B(B)-H (I thought that was what I was doing in [f6fb01a](https://github.com/osrf/buoy_sim/pull/107/commits/f6fb01a6d686394fe13a63079b62e3af1efe306a) in [this line](https://github.com/osrf/buoy_sim/pull/107/commits/f6fb01a6d686394fe13a63079b62e3af1efe306a#diff-a57bd4570256f90b833bc49acbf1b9830e496cec83981c50f5377f558110dc5bR184), but that was giving NaNs. Maybe I wasn't returning what I thought I was returning?

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,203 @@
...
+  Eigen::HybridNonLinearSolver<CatenaryHSoln> catenarySolver(this->dataPtr->catenarySoln);
+  // Tolerance for error between two consecutive iterations
+  catenarySolver.parameters.xtol = 0.0001;
+  // Max number of calls to the function
+  catenarySolver.parameters.maxfev = 1000;
+
+  // Initial guess B = L - V - b
+  // TODO Is this how to pass in L - V - b as initial guess for B?
+  double b = 0.0;
+  this->dataPtr->B[0] = this->dataPtr->L - this->dataPtr->V - b;
```
andermi, line 161: B[0] also needs to be updated like this inside the while loop (after b is incremented... ). Currently, the initial guess for B never changes

buoy_gazebo/src/MooringForce/CatenarySoln.hpp
```diff
@@ -0,0 +1,188 @@
...
+
+    // Solver for x. Pass in initial guess.
+    // Want y(H)=V, so y(H) - V = 0
+    // After this, you have independent variable = H, dependent variable = V
+    int solverInfo = vSolver.solveNumericalDiff(x);
+
+    // Once have x (hopefully x=H)
+    fvec.resize(x.size());
+    // Update H, which is the solution of x from solver
+    fvec[0] = x[0];
```
andermi, line 179: this line needs to be:
```
fvec[0] = x[0] - this->H;
```

buoy_gazebo/src/MooringForce/CatenarySoln.hpp
```diff
@@ -0,0 +1,188 @@
...
+  // Know 0 <= B < L - V. Take in B = L - V - b as initial guess
+  int operator()(const Eigen::VectorXd & B, Eigen::VectorXd & fvec) const
+  {
+    if (B.size() < 1) {
+      ignerr << "Invalid input size for CatenaryHSoln::operator()" << std::endl;
+      return -1;
+    }
+
+    // Scaling factor
+    double c = CatenaryFunction::CatenaryScalingFactor(
```
andermi, line 154: I don't think you need `c` here

buoy_gazebo/src/MooringForce/MooringForce.cpp
andermi: actually, you can just move this down there after b += 1.0
It doesn't need to be done here with b=0 since B is strictly less than L - V.

andermi:
> but that was giving NaNs

possibly because the initial guess of `B = L - V - 0` is invalid since B is strictly less than L - V

> Or should it be returning y_B(B)-H

this ^^. it should be finding the B that makes `x-H == 0`. Just returning x tries to make `x == 0` given a B which will obviously not give `x == H` (unless the buoy is directly over the anchor, in which case the assumptions of this catenary calc are invalid)

the assumptions are that we can use this calculation when the mooring line is pulled sufficiently taut such that the section of the line on the floor, B, is basically pulled straight with no loops/bights. Also, the catenary equation is only valid when H > B, otherwise the line just makes a straight line starting from B up to the buoy. This all means there's some radius around the anchor where we can start to use this solution.

buoy_gazebo/src/MooringForce/CatenarySoln.hpp
mabelzhang: Thanks! a7b2aeeb0784ab488c028136444d3247d009a969

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: a7b2aeeb0784ab488c028136444d3247d009a969
I ended up putting it at the end of the loop, because otherwise the solverInfo ends up 32767 for the given V and H after the entire loop. If I let it call the solver for b=0, then solverInfo actually gets to 1 and the loop breaks out. Seems more right because for those values, it shouldn't never find a solution...

buoy_gazebo/src/MooringForce/CatenarySoln.hpp
mabelzhang: a7b2aeeb0784ab488c028136444d3247d009a969

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Let me know if you think we shouldn't evaluate at b = 0 at all. The Python code evaluates b = 0. So the current code matches that better - evaluates at b = 0 before incrementing b.

mabelzhang:
I think we're getting somewhere.
a7b2aee addresses PR feedback above - thanks for the math corrections! Gosh that makes more sense.
The solverInfo looks better, getting to `1` which I think means it found a solution:
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:190] HSolver solverInfo: 1 c: 0 B: 78 Tx: 0 Ty: -1640
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:190] HSolver solverInfo: 1 c: 0 B: 78 Tx: 0 Ty: -1640
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:190] HSolver solverInfo: 1 c: 0 B: 78 Tx: 0 Ty: -1640
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:190] HSolver solverInfo: 1 c: 0 B: 78 Tx: 0 Ty: -1640
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:190] HSolver solverInfo: 1 c: 0 B: 78 Tx: 0 Ty: -1640
```

So I went ahead and in 4c35cd3 updated V and H (vertical and horizontal distances between buoy and anchor) live, based on the latest buoy position and the virtual anchor position (defined in SDF plugin parameter `anchor_position`).
Outputs look reasonable to me:
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.4321 B: 120.027 Tx: -528.643 Ty: -799.457
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.4559 B: 120.073 Tx: -529.119 Ty: -798.533
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.4367 B: 120.039 Tx: -528.733 Ty: -799.213
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.3683 B: 119.914 Tx: -527.366 Ty: -801.715
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.4136 B: 120 Tx: -528.271 Ty: -799.996
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.3753 B: 119.931 Tx: -527.505 Ty: -801.382
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.4698 B: 120.109 Tx: -529.397 Ty: -797.827
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.397 B: 119.975 Tx: -527.94 Ty: -800.5
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:260] HSolver solverInfo: 1 c: -26.36 B: 119.908 Tx: -527.199 Ty: -801.838
```

If this looks reasonable to you, the next step is to update the test.

buoy_gazebo/src/MooringForce/MooringForce.cpp
andermi: Yeah the python code is also wrong with this.

Shouldn't evaluate with `b = 0` since that will make `c = 0` and should have caused a divide-by-zero error in the `cosh` function.
```
c = ((_L - _B)^2 - _V^2) / (2 * _V)
c = (_L - _L - _V - b)^2 - _V^2
c = _V^2 - _V^2
c = 0
...
y = c * cosh((x[0] - this->B) / c) - c
y = 0 * cosh((x[0] - this->B) / 0) - 0
```

So, the divide by 0 is bad and multiplying the whole `cosh` by 0 is also bad. Also, c should be greater than 0

andermi:
see my new comment in the code above. Also, c should be greater than 0 and your print shows negative. In the python code (with the change to not evaluate at `b=0`) the solution to match is:

given:
```
w = 20
L = 160
H = 120
V = 82
```

results:
```
B = 35.74
c = 53.16
F_x = -1063.14
F_y = -2485.30
S = 124.26
```

buoy_gazebo/src/MooringForce/MooringForce.cpp
andermi: in the while loop, I also just changed it to check for:

```
B_init = L - V - b
if ((L - B_init)**2. - V**2.) <= 0.:
    continue
```

andermi:
[mooring_catenary_calc.zip](https://github.com/osrf/buoy_sim/files/10528134/mooring_catenary_calc.zip)
Try this new calc

mabelzhang:
Sorry about the delay. Ran around doing errands for a few days and had to make up work on a few nights this week.

With V and H hardcoded and fixed (comment out the two calls to `UpdateVH()` in MooringForce.cpp), the numbers are matching the new Python script:
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 82 H: 120 c: 53.1572 B: 35.735 Tx: 1063.14 Ty: -2485.3
```

I realized what made c negative before. I had the anchor_position SDF parameter set to 20, 0, -77, which made H very small, 20. In the Python plot, that shows up off screen:

2023-02-04-054552_1723x1230_scrot

Dragging H there in the Python GUI passed the vertical point when it is 0, which is when the input is invalid.
However, the Python script was still able to come up with a solution. Running the Python script with one of the (V, H) pairs in the output above
```
V = 75
H = 20
```
gives this:
```
y = 21.05*np.cosh((x-66.29)/21.05) - 21.05
B = 66.29 (m)
c = 21.05 (m)
F_x (buoy) = -420.99 (N)
F_y (bouy) = -1874.29 (N)
F_x (anchor) = 420.99 (N)
S = 93.71 (m)
```
whereas in the Gazebo plugin (with `UpdateVH()` uncommented), it's getting negative c:
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 75 H: 20 c: -26.8093 B: 119.955 Tx: -536.187 Ty: -800.9
```

Is the input H = 20 invalid? What does it mean when it's off the screen in the Python GUI?
Maybe the C++ solver is passing the H = 0 point when it's trying to find a solution, and that's why it's stuck somewhere invalid with a negative c?
I don't know. Any insights?

This doesn't fix what c is at those values, but I've changed the initial anchor position to be close to the default values in the Python script:
```
      <anchor_position>120 0 -84</anchor_position>
```
After running for 4+ minutes, c is still stable:
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.5472 H: 139.414 c: 53.9064 B: 35.735 Tx: 1078.13 Ty: -2485.3
```
The effect in the world, with Tx and Ty applied, is that the buoy is drifting toward -x and -y.
2023-02-04-060020_2000x1690_scrot

If I move the buoy manually (pause, change the pose numbers, unpause) to... I think I put (-10, 5, 0) maybe, H shows an update, and c is still okay:
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.5466 H: 110.537 c: 53.9075 B: 35.735 Tx: 1078.15 Ty: -2485.3
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.5466 H: 110.537 c: 53.9075 B: 35.735 Tx: 1078.15 Ty: -2485.3
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.5466 H: 110.537 c: 53.9075 B: 35.735 Tx: 1078.15 Ty: -2485.3
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.55 H: 130.096 c: 53.9018 B: 35.735 Tx: 1078.04 Ty: -2485.3
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.55 H: 130.096 c: 53.9018 B: 35.735 Tx: 1078.04 Ty: -2485.3
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:261] HSolver solverInfo: 1 V: 81.55 H: 130.097 c: 53.9018 B: 35.735 Tx: 1078.04 Ty: -2485.3
```

I don't know. What do you think? Should we manually move it to more positions to test it, should we be concerned that it doesn't work for some numbers - do you think it would ever naturally drift to those values where it's invalid (anchor at 20, 0, -77)?

andermi:
So, in the python calc, I didn't code up all the sanity checking we would need to do for this solution.  This solution only really works when H > B. To implement this solver, I would suggest we not care about when H < B since, in that case, we are sufficiently close to the anchor and not worried about drifting away. This would mean we wouldn't care about solving the catenary inside a certain radius. Once we are outside the radius, we can start solving for the catenary and the forces to move the buoy back home. I would use polar coordinates to solve this and making use of the radial symmetry of the problem. What I mean is, find the radius we are away from the achor and that is our H. If we have (x,y,z) coords, V = z and H = r = sqrt(x^2 + y^2). If that radius is too small such that H < B, we don't care about the solution.

andermi:
Oh and then the `F_x` force to move the buoy back toward the anchor would really be `F_r` and would need to be converted back to (Fx, Fy) making use of polar to cartesian conversion. So in that case, also keep track of `theta` from the cartesian to polar conversion when you find H = r

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,281 @@
...
+
+  // Update vertical distance between buoy and anchor
+  this->V = fabs(this->buoyPos[2] - this->anchorPos[2]);
+
+  // Update horizontal distance between buoy and anchor
+  this->H = sqrt(
+    (this->buoyPos[0] - this->anchorPos[0]) *
+    (this->buoyPos[0] - this->anchorPos[0]) +
+    (this->buoyPos[1] - this->anchorPos[1]) *
+    (this->buoyPos[1] - this->anchorPos[1]));
```
andermi, line 160: So, you're already finding H = radius from anchor? I just noticed this. Should also grab
```
this->theta = atan2(this->buoyPos[1] - this->anchorPos[1], this->buoyPos[0] - this->anchorPos[0]);
```
so you can convert the catenary force to x,y at the end... see below

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,281 @@
...
+    // Solve for B, pass in initial guess
+    solverInfo = catenarySolver.solveNumericalDiff(this->dataPtr->B);
+
+    // Recalculate c with newly solved B
+    c = CatenaryFunction::CatenaryScalingFactor(
+      this->dataPtr->V, this->dataPtr->B[0], this->dataPtr->L);
+
+    // Found solution
+    if (solverInfo == 1)
+      break;
```
andermi, line 252: Don't break... we need to wait for c > 1e-5

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,281 @@
...
+  }
+
+  // Horizontal component of chain tension, in Newtons
+  // Force at buoy heave cone is Fx = -Tx
+  double Tx = c * this->dataPtr->w;
+  // Vertical component of chain tension at buoy heave cone, in Newtons
+  double Ty = - this->dataPtr->w * (this->dataPtr->L - this->dataPtr->B[0]);
```
andermi, line 259: So, the `Tx` and `Ty` in the python calc is badly labeled and should be `T_r` and `T_z` respectively. Using the `this->theta` from above,
```
T_r = - c * this->dataPtr->w;
T_x = T_r * cos(this->theta);
T_y = T_r * sin(this->theta);
T_z = - this->dataPtr->w * (this->dataPtr->L - this->dataPtr->B[0]);
```

so, then below:
```
gz::math::Vector3d force(T_x, T_y, 0.0);  // But, let's just leave out the weight of the mooring chain for now...
```

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Should we add a guard to not apply the force if `solverInfo != 1`?

mabelzhang:
Thanks for looking through! I've updated the PR (2829ad0).

1. An `effectiveRadius` in meters is added, which can be passed in through the SDF parameter:
   ```
      <enable_beyond_radius>50</enable_beyond_radius>
   ```
   When H is below this, the plugin won't do anything.

   Test with
   ```
      <anchor_position>20 0 -84</anchor_position>
   ```
   To see that the plugin is not doing anything, in the current hash, simply check that no debug messages are printed.

2. With the update in Tx and Ty, I changed the anchor position so that it'd have a non-zero y, to see a nonzero Ty:
   ```
      <anchor_position>80 100 -84</anchor_position>
   ```

   That gives a slightly larger Ty than Tx as expected:
   ```
   [ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:286] HSolver solverInfo: 1 V: 81.3665 H: 127.916 c: 54.2071 B: 35.735 Tx: 677.27 Ty: 846.562
   ```

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: 2829ad0

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Added a guard in 2829ad0 to not apply any forces

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Done in 2829ad0. Thanks!

Is H > 50 a reasonable threshold (added as `effectiveRadius`) to enable the mooring force?

andermi:
Something is still not quite right... c and B aren't changing as the buoy drifts. It's just a constant:
`c: 54.2071 B: 35.735`
B should be increasing as the buoy moves toward the anchor

buoy_gazebo/src/MooringForce/MooringForce.cpp
andermi: This still needs to be updated. Using `b=0` and using `B = L - V - 0` will produce a bad `c`

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,307 @@
...
+  catenarySolver.parameters.xtol = 0.0001;
+  // Max number of calls to the function
+  catenarySolver.parameters.maxfev = 1000;
+
+  // Initial guess B = L - V - b
+  double b = 0.0;
+  this->dataPtr->B[0] = this->dataPtr->L - this->dataPtr->V - b;
+
+  // Scaling factor
+  double c = CatenaryFunction::CatenaryScalingFactor(
```
andermi, line 256: the `c` here is going to be bogus with the initial `B` using `b=0`

andermi: just delete setting B above, leave b=0, and set c=0 and let the loop run.

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,307 @@
...
+    this->dataPtr->V, this->dataPtr->B[0], this->dataPtr->L);
+
+  int solverInfo;
+  while ((c <= 1e-5) && (this->dataPtr->L - this->dataPtr->V - b > 0.0)) {
+    // Increment b, to iterate on B
+    b += 1.0;
+    this->dataPtr->B[0] = this->dataPtr->L - this->dataPtr->V - b;
```
andermi, line 263: after computing this, can do a sanity check:
```
if ((L - B)**2. - V**2.) <= 0.:
    continue
```

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,307 @@
...
+  // Horizontal component of chain tension, in Newtons
+  // Force at buoy heave cone is Fx = -Tx
+  double Tr = - c * this->dataPtr->w;
+  double Tx = Tr * cos(this->dataPtr->theta);
+  double Ty = Tr * sin(this->dataPtr->theta);
+  // Vertical component of chain tension at buoy heave cone, in Newtons.
+  // Unused at the moment
+  // double Tz = - this->dataPtr->w * (this->dataPtr->L - this->dataPtr->B[0]);
+
+  igndbg << "HSolver solverInfo: " << solverInfo
```
andermi, line 287: Could you add printing Tr, theta, and Tz?

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: b32e2b9

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Should we just start at `b=1` then? Because c is in the condition, it has to be computed before entering the loop. Then update b at the end of the loop? b32e2b9

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: b32e2b9

mabelzhang:
Changed loop mechanics (b32e2b9). Now starting at `b=1` instead of `b=0`.

Now B is changing - hovering between 77.0000 and 77.5183, e.g.
```
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4831 H: 128.062 b: 1 B: 77.5169 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.483 H: 128.062 b: 1 B: 77.517 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.483 H: 128.062 b: 1 B: 77.517 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4829 H: 128.062 b: 1 B: 77.5171 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4829 H: 128.062 b: 1 B: 77.5171 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4828 H: 128.062 b: 1 B: 77.5172 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4828 H: 128.062 b: 1 B: 77.5172 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.66
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4827 H: 128.062 b: 1 B: 77.5173 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4827 H: 128.062 b: 1 B: 77.5173 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4826 H: 128.062 b: 1 B: 77.5174 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4826 H: 128.062 b: 1 B: 77.5174 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4825 H: 128.062 b: 1 B: 77.5175 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4825 H: 128.062 b: 1 B: 77.5175 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4824 H: 128.062 b: 1 B: 77.5176 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4824 H: 128.062 b: 1 B: 77.5176 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4824 H: 128.062 b: 1 B: 77.5176 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4823 H: 128.062 b: 1 B: 77.5177 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4823 H: 128.062 b: 1 B: 77.5177 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4823 H: 128.062 b: 1 B: 77.5177 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.65
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4822 H: 128.062 b: 1 B: 77.5178 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4822 H: 128.062 b: 1 B: 77.5178 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4821 H: 128.062 b: 1 B: 77.5179 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4821 H: 128.062 b: 1 B: 77.5179 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4821 H: 128.062 b: 1 B: 77.5179 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.4821 H: 128.062 b: 1 B: 77.5179 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.482 H: 128.062 b: 1 B: 77.518 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.482 H: 128.062 b: 1 B: 77.518 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.64
[ruby $(which gz) sim-1] [Dbg] [MooringForce.cpp:293] HSolver solverInfo: 21972 V: 81.482 H: 128.062 b: 1 B: 77.518 c: 1.00614 theta: -2.24554 Tx: 12.5706 Ty: 15.7132 Tr: -20.1227 Tz: -1649.6
```

andermi:
Cool. Now I think we need to convert the forces to Newtons... I can't recall what units it was in the python solution and I can't check right now.

andermi:
> convert the forces to Newtons

I figured out how to look on my phone lol... Hmmm `w * c` is already in Newtons. Those numbers seem small. Can you check those numbers against the python code? Tx in python is -Tr

andermi:
> solverInfo: 21972

just noticed this... I've never seen that number!

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -0,0 +1,322 @@
...
+
+    // Solve for B, pass in initial guess
+    solverInfo = catenarySolver.solveNumericalDiff(this->dataPtr->B);
+
+    // Recalculate c with newly solved B
+    c = CatenaryFunction::CatenaryScalingFactor(
+      this->dataPtr->V, this->dataPtr->B[0], this->dataPtr->L);
+
+    // Increment b, to iterate on B
+    b += 1.0;
```
andermi, line 281: This is a little more readable imho and matches the python code
```suggestion
  double b = 0.0;

  // Catenary scaling factor
  double c = 0.0; 

  int solverInfo;
  while ((c <= 1e-5) && (this->dataPtr->L - this->dataPtr->V - b > 0.0)) {
    // Increment b, to iterate on B
    b += 1.0;

    // Update guess for B, with new b
    this->dataPtr->B[0] = this->dataPtr->L - this->dataPtr->V - b;

    // Invalid chain length. One side of triangle is negative length.
    if ((this->dataPtr->L - this->dataPtr->B[0]) *
        (this->dataPtr->L - this->dataPtr->B[0]) -
        this->dataPtr->V * this->dataPtr->V <= 0) {
      continue;
    }

    // Solve for B, pass in initial guess
    solverInfo = catenarySolver.solveNumericalDiff(this->dataPtr->B);

    // Recalculate c with newly solved B
    c = CatenaryFunction::CatenaryScalingFactor(
      this->dataPtr->V, this->dataPtr->B[0], this->dataPtr->L);
```

andermi:
> > solverInfo: 21972
> 
> just noticed this... I've never seen that number!

```
     enum Status {
         Running = -1,
         ImproperInputParameters = 0,
         RelativeErrorTooSmall = 1,
         TooManyFunctionEvaluation = 2,
         TolTooSmall = 3,
         NotMakingProgressJacobian = 4,
         NotMakingProgressIterations = 5,
         UserAsked = 6
     };
```
These are the possible outputs of the solver... I'm not sure why you are getting those other numbers unless the solver is never running in the first place which would mean we are never running the while loop.

andermi:
@mabelzhang please see https://github.com/osrf/buoy_sim/pull/129 for some updates

mabelzhang:
Thank you for the fixes! I'm seeing good values.

Performance is still bad. On `main`, I get 40-60% RTF, whereas this PR is running at 4-6%.
I can try messing with the solver parameters?

I've updated the test (anchor position, V, L, drift force, number of iterations).
With the virtual mooring, it doesn't crash to a huge force being applied anymore, which is good.
Comment out the `gzdbg` line in MooringForce.cpp for a faster run (will comment out permanently before merging PR):
```
$ ./build/buoy_tests/mooring_force
```

So, this is amusing. To stress test the mooring force, I've made the drift force really big, and increased the iterations until the distance value approaches an asymptote:
Originally, I set as test condition the max horizontal distance formed by the triangle of sqrt(L^2 - V^2), which is about 137 m. Turns out the horizontal distance between the buoy and anchor actually exceeds 137 and eventually approaches 160, but never gets to 160.
160 is L, the length of the mooring chain. That can only happen if the buoy and the anchor are at the same depth, and checking the z value of the buoy shows that the buoy is basically getting pulled all the way to the seafloor (-77 m) :laughing:
```
      gz::math::Vector3d force(1000000, 0, 0);
```
```
  int targetIterations{20000};
```
```
[Dbg] [MooringForce.cpp:294] HSolver solverInfo: 1 V: 6.22962 H: 159.965 b: 1 B: -586.85 c: 44765.7 theta: -0.333468 Tx: -845993 Ty: 293056 Tr: -895313 Tz: -14937
[Dbg] [mooring_force.cpp:116] Final buoy link pose: 231.156 47.6463 -77.77 Final distance between buoy and anchor: 159.965
```

I interpret that result as, the mooring force is working, since the difference in distance gets really close but never exceeds L=160.

The force being applied is extraordinarily large (1 million) and probably will never happen in real life, so I'm not concerned about the buoy getting pulled to the seafloor. We don't have a bound on the mooring force. We're not using the anchor's weight, no Tz is being applied, so I don't know how it goes to the bottom. But even Tr is huge, and maybe that eventually builds up something in the vertical component.

The test will run faster once we fix performance.

andermi:
hah! that is awesome!

So, yeah, I'm sure we could reduce the max number of iterations and the solution tolerance of both solvers to speed up the result. We aren't really going for accuracy here, just a restoring force back to the anchor.

mabelzhang:
848e371 tweaks the solver's `maxfev` and `xtol` to increase performance a bit... 24-28% compared to 60% on `main`... Still not that great but that's about the limit, unless we want to increase the test tolerance by a LOT (like 50 m over 160 m). Details below.

Performance:
`main` runs at RTF 50-60% for me.
With `maxfev=1000` and `xtol=0.0001` for both solvers, I got RTF 4-6% (original setting above).

With `maxfev=100` and `xtol=0.0001` for both solvers, I got RTF 14-16%.
With `maxfev=100` and `xtol=0.001` for both solvers, I got RTF 18-20%. Test passes.
With `maxfev=100` and `xtol=0.01` for both solvers, I got RTF 4-10% and it crashed quickly (tried twice, both crashed).
With `maxfev=50` and `xtol=0.001` for both solvers, I got RTF 19-22%. Test passes.
With `maxfev=50` and `xtol=0.01` for both solvers, I got RTF 19-22%. Test explodes.
With `maxfev=20` and `xtol=0.001` for both solvers, I got RTF 24-28%. Test passes.
With `maxfev=20` and `xtol=0.01` for both solvers, I got RTF 26-28%. Test explodes.
With `maxfev=20` and `xtol=0.1` for both solvers, I got RTF 34-38%.
With `maxfev=10` and `xtol=0.001` for both solvers, I got RTF 30-34%. Test fails.
With `maxfev=10` and `xtol=0.01` for both solvers, I got RTF 30-34%. Test fails.
With `maxfev=10` and `xtol=0.1` for both solvers, I got RTF 34-38%. Test fails and explodes.

Some of those values are probably too inaccurate, as they cause the test to explode.

`10` and `0.1` causes test to fail and explode:
```
[Dbg] [MooringForce.cpp:295] HSolver solverInfo: 1 V: 5.00741 H: 171.624 b: 1 B: 154.909 c: 0.084382 theta: -0.613322 Tx: -1.38005 Ty: 0.971384 Tr: -1.68764 Tz: -101.822
/home/developer/mbari/buoy_sim_ws/src/buoy_sim/buoy_tests/tests/mooring_force.cpp:104: Failure
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 174.24 vs 160
[Dbg] [MooringForce.cpp:295] HSolver solverInfo: 1 V: 4.57022 H: 174.24 b: 1 B: 155.34 c: 0.0904647 theta: -0.634121 Tx: -1.45755 Ty: 1.07195 Tr: -1.80929 Tz: -93.1961
/home/developer/mbari/buoy_sim_ws/src/buoy_sim/buoy_tests/tests/mooring_force.cpp:104: Failure
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 1.22099e+07 vs 160
[Dbg] [MooringForce.cpp:295] HSolver solverInfo: 21980 V: 1.01898e+06 H: 1.22099e+07 b: 0 B: 155.34 c: 0 theta: -1.82327 Tx: 0 Ty: 0 Tr: -0 Tz: -93.1961
/home/developer/mbari/buoy_sim_ws/src/buoy_sim/buoy_tests/tests/mooring_force.cpp:104: Failure
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 5.6562e+17 vs 160
[Dbg] [MooringForce.cpp:295] HSolver solverInfo: 21980 V: 4.32243e+17 H: 5.6562e+17 b: 0 B: 155.34 c: 0 theta: 1.93448 Tx: 0 Ty: -0 Tr: -0 Tz: -93.1961
```

`10` and `0.01` fails:
```
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 211.596 vs 160
```
`10` and `0.001` fails:
```
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 208.716 vs 160
```
`20` and `0.01` explodes:
```
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 22572.5 vs 160
```
`50` and `0.01` explodes:
```
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 4254.48 vs 160
```
`100` and `0.01` explodes:
```
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 228.168 vs 160
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 91906.7 vs 160
```

`100` and `0.001` passes:
```
[Dbg] [mooring_force.cpp:116] Final buoy link pose: 231.129 47.5695 -77.7825 Final horizontal distance between buoy and anchor: 159.965
```
`50` and `0.001` passes:
```
[Dbg] [mooring_force.cpp:116] Final buoy link pose: 229.544 43.2027 -77.9807 Final horizontal distance between buoy and anchor: 159.967
```
`20` and `0.001` passes:
```
[Dbg] [mooring_force.cpp:116] Final buoy link pose: 229.222 42.365 -77.838 Final horizontal distance between buoy and anchor: 159.965
```

`20` and `0.001` (24-28% RTF) seems the least tolerant numbers that still pass the test.
For the ones that just fail and don't explode, we can also increase the tolerance of the test, like `10` and `0.01` (30-34% RTF) is 50 meters over 160 m. Seems a lot though.

andermi:
another thing to try is to see how often we are looping in the main loop... how high does `b` get on the original settings? We might be able to come up with a better guess for `B` to reduce the number of guesses.

andermi:
with our radius assumption, we know that `B` will never be more than `H` so, we could start our initial guess at `B = H - b` instead of `B = L - V - b`

andermi:
If we can't get anything to be fast enough, we may have to punt (hey, it's a superbowl pun!) and just do some force proportional to the radius

andermi:
The other crazy idea (I'd prefer the force proportional to radius over this idea...) would be to try throwing the solver loop into a thread, and we could get delayed forces at some rate whenever they're ready... in order to smooth out the async and sporadic forces, we could throw it in a smoothing filter like a FIR. None of this has to be all that accurate.

andermi:
I checked the timing in the python solver and it's usually 2-5 milliseconds to solve. Can you time the whole solve loop in the plugin?

Also, don't do `B = H - b` ... it's much worse of a guess

andermi:
So, I sped up the calc by removing one of the solvers... I realized we can still solve for the right B if we solve for `y_B(B)-V` when `y_B(B)` returns `y(H)` instead of solving for the `x` that makes `y(x) == V`

```
def y_B(B):
        y, _ = catenary(V, B, L)
        return y(H)  # return this now...
```

and solve for this:
```
B = spo.fsolve(lambda B: y_B(B)-V, B_init)[0]
```
(Note: this is different than the original `B = spo.fsolve(lambda B: y_B(B)-H, B_init)[0]`)

andermi:
even using a second order taylor series (bad approx) of cosh, I can't get get the computation down below 1ms

mabelzhang:
6dccd1d does (I think) what we said in the meeting:
Removes inner solver and just uses y(H) (instead of y(x), solving for x with x0=H).
In outer loop, compare y_B(B) to V.

That seems to have sped up RTF for me by 10-15%. Without the commit, I'm getting 18-20%. With the commit, 30-33%.

Will add profiling.

srmainwaring:
I think you may be able to keep the accuracy as the inner loop can be solved in closed form. The snippet below can be dropped into the nice Python example from https://github.com/osrf/buoy_sim/pull/107#issuecomment-1407472828. I've used an initial guess for the solver using a lower bound on B, solver takes about 8 iterations.  

```python
def solve_catenary2(V, H, L):
    # Given the expression for c, x can be calculated in closed form
    # by inverting the catenary equation.
    def y_B(B):
        c = ((L-B)**2. - V**2.)/(2.*V)
        x = c * np.arccosh(V / c + 1.0) + B
        return x

    # Assuming V and H are such that V^2 + H^2 < L^2,
    # A minimum value of B is given assuming the suspended section
    # of the chain S is a straight line from (B, 0) to (H, V).
    # B should not be greater than H (when the chain is vertical).
    def b_min(V, H, L):
        bmin = (L*L - (V*V + H*H)) / (2 * (L - H))
        return min(bmin, H)

    B_init = b_min(V, H, L)
    
    # solve for the B that makes y(H)==V, so y_B(B)==H
    res = spo.fsolve(lambda B: y_B(B)-H, B_init, full_output=True)
    B = res[0][0]
    
#     print(res)
    y, c = catenary(V, B, L)
    return y, B, c
```

mabelzhang:
The closed form makes sense. I've wondered about that :sweat_smile:
I'll take a closer look next week.
`b_min` is the lower bound on B?
Regardless of the initial guess for B, you're saying that either way, the inner loop can be done using the closed form, but we need to play with the initial guess?

mabelzhang:
Profiling:
The entire solver loop, in the state it is now (72a08ba), takes 0.1-0.16 ms (remove the debug printout).
2023-02-18-013058_3840x2160_scrot

srmainwaring:
> `b_min` is the lower bound on B?

No, I was wrong about that. It's the value of B that you get if the chain follows a hockey stick profile: flat along the x-axis to B then a straight line to the buoy. This is a good initial guess but it's not necessarily a minimum for B. The catenary allocates more chain to the section (L - B) which moves the point B right, but the tension lifts more chain off the floor moving the point B left - and it's not clear to me at the moment which one wins. In any case for a gradient solver it seems to be a good initial guess. If the root can be bracketed something like brentq may be better / faster for a scalar function like this. 

> Regardless of the initial guess for B, you're saying that either way, the inner loop can be done using the closed form

Yes. The caveat is that that value c needs to be non-zero (which is satisfied by B < L - V). 

> but we need to play with the initial guess?

I think `b_min` is probably a good guess - but could do with some more analysis. The best check would be a bracket to check that the target function changes sign inside the search region.

srmainwaring:
Checked the proposed changes here: https://github.com/srmainwaring/buoy_sim/commit/66631ea75319b517ba7813bf0850ae7685a4c9f9.

There's a single call to the solver each `PreUpdate` which converges in 8-10 steps. It doesn't seem to need the while loop - but that may just be valid for the provided example?

```bash
[Dbg] [MooringForce.cpp:310] HSolver solverInfo: 1
 V: 81.5613
 H: 123.968
 b: 49.673
 B: 25.9646
 c: 69.3542
 theta: -2.24447
 Tx: 865.353
 Ty: 1084.05
 Tr: -1387.08
 Tz: -2680.71
 nfev: 9
 iter: 8
 fnorm: 4.26326e-14
```

I haven't got the GZ remotery profiler running on macOS, but a check with macOS Instruments has the MooringForce down to < 1% of the SimulationRunner::UpdateSystems (for comparison Buoyancy is 17% and Hydro 2%).

mabelzhang:
Hmm @srmainwaring I tried out your [branch](https://github.com/srmainwaring/buoy_sim/tree/srmainwaring/mooring) with the commit 66631ea, and the behavior is very different - the buoy quickly drifts to +x +y within a few seconds, and continues to drift that way. Is that what you're seeing?

The previous behavior holds the buoy in the same position for quite a while (I ran for sim time 20 s, and the buoy does not drift noticeably at all).

Do you know why that difference might be?

srmainwaring:
@mabelzhang - I don't see that behaviour - it looks like the solver did not converge in your run. In my example I disabled all the plugins except hydrodynamics, buoyancy and mooring.

I've added some extra checks and debug info here: 7e8f4a5

The debug prints at the end of PreUpdate show the solver converging:

```bash
gz topic -e -t /model/MBARI_WEC_ROS/odometry
``` 

Initial forces:

```bash
[Dbg] [MooringForce.cpp:331] HSolver solverInfo: 1
 t: 1
 Anchor: 80 100 -84
 Buoy:   0.300082 0.359317 -2.57162
 R: 90
 L: 160
 V: 81.4284
 H: 127.594
 b: 41.491
 B: 13.766
 c: 90.5938
 theta: -2.24546
 Tx: 1131.76
 Ty: 1414.93
 Tr: -1811.88
 Tz: -2924.68
 nfev: 7
 iter: 6
 fnorm: 3.2987e-07
```

Last update before moving inside the effective radius:

```bash
[Dbg] [MooringForce.cpp:331] HSolver solverInfo: 1
 t: 92
 Anchor: 80 100 -84
 Buoy:   23.951 29.5351 -2.46002
 R: 90
 L: 160
 V: 81.54
 H: 90.0377
 b: 77.502
 B: 74.3112
 c: 4.25434
 theta: -2.24274
 Tx: 52.967
 Ty: 66.5903
 Tr: -85.0869
 Tz: -1713.78
 nfev: 6
 iter: 5
 fnorm: 0.000146596

[Msg] Buoy horizontal radius H [89.9999] is inside effective radius R [90] skipping force update.
```

mabelzhang:
That makes sense. I've forgotten that the mooring is in +x +y.. Thanks for adding the debug info!

@andermi reminded me that as far as #129, buoy did drift toward +x +y (which I had also seen in that PR but had forgotten).
Up to bd42edb, it still drifted.

Then my next commit, 848e371 tweatking xtol and maxfev for performance, broke things and I didn't notice. Guess that performance improvement didn't work X_X

@srmainwaring could you make a PR against this PR with your branch? That way you show up as the contributor and I'm not copy-pasting your code. Thank you!

mabelzhang:
@andermi after #135 is merged, could you point me to (or open a PR targeting this one) how to disable mooring by default in the launch file / parameters? It'll have to go in with this PR. Thanks!

mabelzhang:
Tests:
Locally, up to commit 959fa9e (last non-merge commit to #135), the new mooring test is passing.
From 8500a8f (merge from main), the new mooring test is failing:
```
/home/developer/mbari/mbari_wec_ws/src/mbari_wec_gz/buoy_tests/tests/mooring_force.cpp:104: Failure
Expected: ((buoyXY - anchorXY).Length()) < (maxRadius), actual: 1.75467e+11 vs 160

...

[Err] [MooringForce.cpp:344] HSolver failed to converge, solverInfo: 5 No mooring force will be applied.

ODE INTERNAL ERROR 1: assertion "aabbBound >= dMinIntExact && aabbBound < dMaxIntExact" failed in collide() [collision_space.cpp:460]
Aborted (core dumped)
```
This is the same error as the first ones we saw in this PR (we had fixed it above), the buoy position is exploding and no longer held within the 160 m radius.
We fixed it before, and it was nicely held within 160 m. Now it's broken again after the merge from `main`.
Probably there has been some new behavior in `main` that is changing how models are moving.
We'll need to fix this again.

andermi:
does it still work with the old effectiveRadius code?

mabelzhang:
Yeah 7e79a40 (right before removing `effectiveRadius`) works. Even 959fa9e that replaces `effectiveRadius` with the new `toleranceL` works.
The test breaks starting from the merge from `main`, which... there's so much new stuff. From the printouts, I can tell at least one new thing is WaveBodyInteraction.cpp, which at least wasn't printing stuff prior to the merge from main.
Maybe the test just needs to be loosened.

mabelzhang:
I'm not sure why one test was missing the log, and the other test was timing out. Hard to see what's happening with a lot of printouts from `WaveBodyInteractions.cpp`. Maybe we should throttle that like Rhys throttled the mooring plugin printouts.

Added SDF parameter to disable plugin by default. Might as well see if the tests pass with the plugin disabled.

mabelzhang:
This is weird. I commented out the mooring test, and the CI is still missing logs. I don't see this happening to other PRs. Have we seen this elsewhere?

andermi:
No, I haven't seen that, although github has been acting strangely the last couple days...

andermi:
I reran the test and I noticed it was running for >5hrs before it was cancelled

mabelzhang:
Yeah I saw that. Locally, there's a lot of printouts from `WaveBodyInteractions.cpp` that obscures the printouts from `MooringForce.cpp`. I'll probably disable the printouts from the former to look into the latter more closely.

andermi:
those printouts are already masked as Debug... did you enable debug printing?

mabelzhang:
I tried levels 1-4 in the first line of the mooring test that sets them, but running the test still prints all the debug printouts. I mean, the mooring printouts are debug too, and I want to see those.

andermi:
osrf/mbari_wec_utils#51 should fix CI test issue

mabelzhang:
aff4053 should fix the test now, and a few bugs I found...

The plugin now has an `<enable>` flag that defaults to `false`, so the mooring is off by default.
The test calls a service to enable the mooring plugin, so that it can be tested.

That way this PR can be merged without affecting other things.
If we have time, a launch parameter to turn it on can be added in a separate PR. If not, it can be turned on by a service call to `/mooring_force/enable`.

mabelzhang:
Wow, CI is finally green.

mabelzhang:
Recap:
- Plugin is disabled by default via the `<enable>false</enable>` in the model.sdf file.
- Plugin can be enabled live by calling `/mooring_force/enable` service. The test does that.
- To do in another PR: Add launch parameter to enable or disable mooring.

andermi (approved):
LGTM! :)

PR #108, opened 2022-11-28
Fix batt curr

hamilton8415:
This is a very simple fix of a constant in the electro-hydraulic PTO that had been set wrong.   I re-established the experiment_comparison tests and they're passing.

Unfort, sc_commands_ros_feedback test if failing, I'm hoping @andermi can take a look and see why that is.  It broke it before too, so if there's something obvious causing problems please let me know.


mjcarroll:
> Unfort, sc_commands_ros_feedback test if failing, I'm hoping @andermi can take a look and see why that is

Appears to be passing in CI?

PR #109, opened 2022-11-28
Suggestions to #103

mjcarroll:
Signed-off-by: Michael Carroll <michael@openrobotics.org>


hamilton8415:
Thanks for doing this Michael.  Looks good to me, but I am going to wait until I'm back (Dec 7th) to merge this simple PR.  The reason is I'm working with the FS_Hydrodynamics library code on my Mac while I'm traveling, and having it compile without dependencies on the ROS/Gazebo is convenient.

PR #120, opened 2023-01-23
Fix flaky PC commands test

andermi:
I changed some asserts to expects which I think was causing it to bomb out early and then timeout when launch_test didn't notice it exited.


andermi:
now I can't get it to fail anymore

mabelzhang (approved):

PR #127, opened 2023-02-03
fix `mbari_wec.launch.py` launch args

andermi:
Inadvertently broke the original `mbari_wec.launch.py` when adding some args from batching


mabelzhang (approved):

hamilton8415 (approved):
Thanks Michael

PR #135, opened 2023-03-08
Update mooring force calculation

srmainwaring:
This PR updates the mooring force calculation to replace one of the solver loops with a closed form expression and an alternative estimate for the initial solver position.

See https://github.com/osrf/buoy_sim/pull/107#issuecomment-1454844684 for details.

In addition to these changes it is recommended to add checks on edge cases (where the solver is not expected to converge), and possible a bracket check, however the Eigen newton solver does not specifically require this.


buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -202,12 +211,36 @@ void MooringForce::Configure(
...
   this->dataPtr->B.resize(1U);
+
+  // debug print throttle, default 1Hz
+  {
+    double rate(1.0);
+    if (_sdf->HasElement("debug_print_rate"))
+    {
+      rate = _sdf->Get<double>(
+        "debug_print_rate", this->dataPtr->L).first;
+      gzdbg << "Debug print rate set to " << this->dataPtr->L
```
mabelzhang, line 236: Typos from copy-paste
```suggestion
        "debug_print_rate", rate).first;
      gzdbg << "Debug print rate set to " << rate
```

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -240,9 +275,19 @@ void MooringForce::PreUpdate(
   this->dataPtr->UpdateVH(_ecm);
   // If buoy is not far enough from anchor for there to be a need to pull it
   // back, no need to apply mooring force
-  if (this->dataPtr->H < this->dataPtr->effectiveRadius) {
+  if (this->dataPtr->H < this->dataPtr->effectiveRadius)
+  {
```
mabelzhang, line 279: So, this repo is following the style that the curly braces go on the same line as the conditions. I don't know why. But this plugin should follow the other plugins. Same for all other new blocks in this PR.

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -240,9 +275,19 @@ void MooringForce::PreUpdate(
...
+    if (!this->dataPtr->notifiedInsideEffectiveRadius)
+    {
+      this->dataPtr->notifiedInsideEffectiveRadius = true;
+      gzmsg << "Buoy horizontal radius H [" << this->dataPtr->H << "]"
+            << " is inside effective radius R ["
```
mabelzhang, line 284: This message actually doesn't sound like anything is wrong, which is opposite of the intention. I realize effectiveRadius is maybe not the best name. I don't remember why I called it that. Maybe it should be ineffectiveRadius, as in, mooring plugin is ineffective within this radius. Then the message could say something like, no mooring force is needed when buoy is close to the anchor.

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -254,34 +299,19 @@ void MooringForce::PreUpdate(
...
-  double c = 0.0;
-
-  int solverInfo;
-  while ((c <= 1e-5) && (this->dataPtr->L - this->dataPtr->V - b > 0.0)) {
-    // Increment b, to iterate on B
-    b += 1.0;
+  // Initial estimate for B (upper bound).
+  auto BMax = [](double V, double H, double L) -> double
+  {
+    return (L*L - (V*V + H*H)) / (2 * (L - H));
```
mabelzhang, line 305: Consistent spacing around operators
```suggestion
    return (L * L - (V * V + H * H)) / (2 * (L - H));
```

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -292,28 +322,53 @@ void MooringForce::PreUpdate(
...
+      << " Tz: " << Tz << "\n"
+      << " nfev: " << catenarySolver.nfev << "\n"
+      << " iter: " << catenarySolver.iter << "\n"
+      << " fnorm: " << catenarySolver.fnorm << "\n"
+      << "\n";
+  }
 
   // Did not find solution. Maybe shouldn't apply a force that doesn't make sense
   if (solverInfo != 1) {
+    gzerr << "HSolver failed to converge, solverInfo: " << solverInfo << "\n";
```
mabelzhang, line 356: A more explicit consequence would be helpful to a user (my suggestion is over 80 chars, feel free to edit)
```suggestion
    gzerr << "HSolver failed to converge, solverInfo: " << solverInfo << " No mooring force will be applied." << std::endl;
```

buoy_gazebo/src/MooringForce/MooringForce.cpp
```diff
@@ -292,28 +322,53 @@ void MooringForce::PreUpdate(
...
+      << " c: " << c << "\n"
+      << " theta: " << this->dataPtr->theta << "\n"
+      << " Tx: " << Tx << "\n"
+      << " Ty: " << Ty << "\n"
+      << " Tr: " << Tr << "\n"
+      << " Tz: " << Tz << "\n"
+      << " nfev: " << catenarySolver.nfev << "\n"
+      << " iter: " << catenarySolver.iter << "\n"
+      << " fnorm: " << catenarySolver.fnorm << "\n"
+      << "\n";
```
mabelzhang, line 351: Usually `std::endl` flushes the stream, `\n` doesn't. It makes a difference here, because the timing of these printouts is important. We usually use `std::endl` at the end of all the printouts.

mabelzhang (commented):
Thanks for the improvements! I've tested it. Some suggestions. I'll mostly trust you on the math :sweat_smile:

srmainwaring:
Hi @mabelzhang, thanks for the review feedback. I've not forgotten about this and will come back with updates. In the meanwhile I've borrowed the code for another project - and have made some more wholesales changes removing the specific dependencies on the buoy and including the chain weight contribution. I think this plugin would be very useful upstreamed to `gz-sim7`. Happy to help with this if you want to follow up?

buoy_gazebo/src/MooringForce/MooringForce.cpp
srmainwaring: Fixed in d6f7287

There are a number of other places where the gz coding standards are not followed - will get to those as well while I'm at it (follow up commit).

buoy_gazebo/src/MooringForce/MooringForce.cpp
srmainwaring: I think we can get rid of the effective radius and replace it with a different value - a tolerance on the radius at which the catenary calculation takes effect. Then you don't need a message here.

In https://github.com/srmainwaring/asv_sim/pull/21 I've changed the calculation to check if H + V <= L + tolerance, where tolerance is about 0.5 m. When this is true the chain will hang vertically, so no horizontal tension and only the weight of the suspended chain is applied (F = w * V (N)). When H + V > L + tolerance the catenary solver is invoked and both vertical and horizontal forces are applied. This seems to work well.

buoy_gazebo/src/MooringForce/MooringForce.cpp
srmainwaring: LGTM.

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Ah, I'm sorry, I actually meant that we should follow the style in this repo - to have braces on the same line. I'm not sure it is the goal of this repo to follow Gazebo coding standards. Last I checked, the plugins tend to have braces on the same line.

srmainwaring: Ok - my misunderstanding. If you're ok to clean up formatting that works for me.

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: > the weight of the suspended chain

We're not actually using the weight `w` right now - `Tz` is unused, by design. Does that still align with your intention for the tolerance?

@andermi FYI this is a change from the 90 m tolerance we previously decided on, to a flexible tolerance as first mentioned here https://github.com/osrf/mbari_wec_gz/pull/107#issuecomment-1459752464

buoy_gazebo/src/MooringForce/MooringForce.cpp
mabelzhang: Yeah I'll do it. I'll unresolve this comment to remind myself after merging

mabelzhang (commented):
Thanks for iterating! Just one question.

srmainwaring:
> We're not actually using the weight w right now - Tz is unused, by design. Does that still align with your intention for the tolerance?

I haven't change the tolerance calculation in the PR, and don't think it is affected by the inclusion / exclusion of the vertical component.

To implement the tolerance changes for my own version I've replaced

```c++
  // Update V and H based on latest buoy position
  this->dataPtr->UpdateVH(_ecm);
  // If buoy is not far enough from anchor for there to be a need to pull it
  // back, no need to apply mooring force
  if (this->dataPtr->H < this->dataPtr->effectiveRadius)
  {
    if (!this->dataPtr->notifiedInsideEffectiveRadius)
    {
      this->dataPtr->notifiedInsideEffectiveRadius = true;
      gzmsg << "Buoy horizontal radius H [" << this->dataPtr->H << "]"
            << " is inside effective radius R ["
            << this->dataPtr->effectiveRadius << "]"
            << " skipping force update." << std::endl;
    }
    return;
  }
  this->dataPtr->notifiedInsideEffectiveRadius = false;
```

with

```c++
  // Update V and H based on latest buoy position
  this->dataPtr->UpdateVH(_ecm);

  // Skip solver if the chain can drop vertically (within tolerance).
  double toleranceL = 0.1;
  if (this->dataPtr->V + this->dataPtr->H <= this->dataPtr->L + toleranceL)
  {
    // Assume all force is vertical.
    double Tz = - this->dataPtr->w * this->dataPtr->V;

    math::Vector3d force(0.0, 0.0, Tz);
    math::Vector3d torque = math::Vector3d::Zero;
    if (force.IsFinite() && torque.IsFinite())
    {
      // this->dataPtr->link.SetVisualizationLabel("Mooring");
      this->dataPtr->link.AddWorldWrench(_ecm, force, torque);
    }
    else
    {
      gzerr << "[Mooring] force is not finite.\n";
    }
    return;
  }
```

When the slack has not been taken up the chain hangs vertically and the buoy is subject to the weight of the chain. As soon as the buoy starts to move further away from the anchor the chain must lift off the seabed a little, and so the horizontal force is also added. The net force will initially be mostly directed down, with the horizontal component increasing as the buoy drifts further from the anchor point. The behaviour is (nearly) continuous in direction and magnitude. The small tolerance is to prevent the solver from failing to converge by asking it to find a catenary solution when the chain is almost vertical. The full implementation is in the link to my `asv_sim` repo if you want to see all the changes.

If you set an arbitrary cutoff the force will change from zero to a possibly large value in one time step. If the buoy is moving back and forth across that cutoff it may introduce some odd signals into the physics, which may or may not be material.  

If you'd like to follow up on why I made the changes and would recommend them I'd be happy to follow up with a call (maybe the Monday meeting next week). I've use them in simulations for 3-4 light buoys (which move a lot) in moderate sea states in simulations that ran for 20 - 30 minutes RT with no convergence issues on the solver.

buoy_gazebo/src/MooringForce/MooringForce.cpp
andermi: Yeah, we are following the ROS 2 linters not GZ

andermi: I suppose if we want to upstream this, perhaps it would be best to split this off to another repo/workspace and follow GZ standards.

buoy_gazebo/src/MooringForce/MooringForce.cpp
srmainwaring: @andermi and @mabelzhang - I suspect I had different requirements / objectives in the modifications made here in addition to the speed up. I was aiming for something that would be physically realistic and included the weight force of the anchor chain - expecting this to modify the behaviour of the moored object. So for example a light mark which would otherwise lie on its side would be held upright when moored with a chain. 

If you are after something that just limits the drift without otherwise affecting the moored object and are not too worried about continuity of the applied forces etc. then the changes suggested are probably not relevant.

andermi: @hamilton8415 has been curious (as we've never studied it) how the mooring might affect energy conversion if it at all. So, we aren't necessarily ready to throw out the weight forever, but modeling that is not on the roadmap in the near term.

srmainwaring: > I do think this would be a worthwhile plugin to upstream with more care toward the general case

I'd be happy to help with that. I think you could also add a GUI plugin to visualise the mooring cable which could be quite nice.

andermi: > visualise the mooring cable

Perhaps you have some tricks up your sleeve here?

My other reason for choosing a radius that the mooring starts to take effect is to avoid the condition where the chain is not taut enough for the solution to be valid (chain coiled up or folding back on itself and catenary not in line with B )

srmainwaring: > avoid the condition where the chain is not taut enough for the solution

I assumed that whenever H + V <= L + some_small_tolerance the chain would be vertical, and the catenary solver is skipped. When H + V >  L the chain must be under tension and the solver is applied. I think that amounts to much the same thing, except the radius is dynamic as V varies with water depth / buoy bobbing about.

srmainwaring: > Perhaps you have some tricks up your sleeve here?

If you add visuals using a GUI plugin they don't need to be attached to links or other model quantities. I was thinking you could add a series of small cylinder visuals following the catenary using the same approach used in the force visualization plugin. You could use markers as well, but the visuals seem to work well and have a bit more flexibility in terms of the shapes available. If I have some time to spare I'll see if I can prototype something.

andermi: > whenever H + V <= L + some_small_tolerance the chain would be vertical, and the catenary solver is skipped. When H + V > L the chain must be under tension and the solver is applied

Yeah that works. In the case of H + V == L, the chain would make an L along H and V. For the `<` case, it would be folding back on itself, or coiled/curved in some way. For the `>` case, it would be taut and starting to pull a catenary

mabelzhang: So is the current code fine as it is to merge, or do we need to make use of `Tz`?

mabelzhang: @srmainwaring Should I go ahead and paste those changes you suggested in your comment, or do you want to do it?

srmainwaring: @mabelzhang - I'm fine for you do to it, as I think this PR merges into your branch and then that goes into main. So if you're ok to copy it in all good.

The only catch is if you add the Tz force from the comment then add it later too (or don't add it in the pasted comment if not later). I think adding the vertical force is the physically correct thing to do, but there may be reasons you don't want that for the wave buoy (calibration, tests etc that assume certain masses, and this would mess that up).

mabelzhang: Okay I've added the suggested change with `toleranceL` (959fa9e).
Since we aren't using Tz, the drop down weight vector (0, 0, Tz) then becomes all 0s, which is basically skipping the force update. That preserves the previous behavior with `effectiveRadius`. I've commented out the Tz block since we aren't using weight, in case we want to use the weight in the future.

mabelzhang (approved):
Thank you so much for the addition!
Will merge from target PR and wait for CI.

mabelzhang:
Gonna merge into target PR and try luck with CI there.
CI is timing out probably due to usual CI machine slowness. This PR is not changing any tests, and the mooring test is passing for me locally.

PR #138, opened 2023-03-20
added FetchContent to CMake files for FreeSurfaceHydrodynamics library

hamilton8415:
I am experimenting with using CMake FetchContent to draw in the FreeSurfaceHydrodynamics source instead of relying on installing a .deb from the ppa. This works nicely, when I try to run the executables for which this is a requirement for directly.  However, when I run those executables from a launch file, it complains of not being able to find the .so library, etc.  

There are a few places this library is used, the WaveBodyInteraction and IncidentWave plugins.  and the heaveonly and pitchonly tests.  In the plugins the CMake file inclusion is straightforward. For the tests it's more complicated and some elements of what needs to be done is hard-coded for testing, and can be improved later.

But, for now, interested if anyone has an idea of why it's not finding the libraries...


andermi:
hmm yeah odd. I would perhaps try it flat without loops or macros and see if you can get it to work, then wrap it back up into macros/loops once it works

mabelzhang:
Looks like GitHub wasn't able to figure out the diffs between fs_hydrodynamics and main when it automatically retargeted to `main`. There are a lot of diffs. Probably needs a manual fast-forward or rebase.

PR #140, opened 2023-04-12
add support for pblog

andermi:
added to batching and standard launch

requires:
https://github.com/osrf/mbari_wec_utils/pull/44 <-- https://github.com/osrf/mbari_wec_utils/pull/47

### DIFF
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -10,7 +10,7 @@ except NameError:
 try:
     inc_wave_seed
 except NameError:
-    inc_wave_seed = 42  # not defined so default
+    inc_wave_seed = 0  # not defined so default to 0 which changes seed every run
 
 # Check if battery state (battery_soc or battery_emf) was passed in via empy
 try:
--- a/buoy_description/models/mbari_wec_base/model.sdf.em
+++ b/buoy_description/models/mbari_wec_base/model.sdf.em
@@ -483,6 +483,25 @@ buoyancy_radius = ((3*buoyancy_disp)/(4*math.pi))**(1/3)
           </box>
         </geometry>
       </collision>
+      <sensor name='tf_imu' type='imu'>
+        <!-- Physical IMU rotated 90 with y/z horizontal -->
+        <pose>0.0 0.0 0.0 0.0 1.570796 0.0</pose>
+        <topic>HeaveCone_link/tf_imu</topic>
+        <update_rate>50</update_rate>
+        <imu>
+          <orientation_reference_frame>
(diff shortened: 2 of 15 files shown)


hamilton8415:
Looking good.  Could consider changing "latest_csv" to "latest" to match buoy behaviors.

andermi:
> consider changing "latest_csv" to "latest"

I can do that, but I was hoping to distinguish between pblog and rosbag. No biggie if you'd like to have `latest` be pointing to the csv and perhaps `latest_rosbag` pointing to the rosbag

hamilton8415:
I'm flexible.  Nothing automated relies on this and it's just a convenience so it can stay as is.

In general, I noticed a few things when I was running this yesterday (I was running from batch mode)
1) When the job started there was immediately a .gz version of the .csv file, even though the .csv file was still growing.  Same filename.
2) When it completed, the "latest" file pointed to the .gz file.  That is OK, but not really necessary.  For me anyway, I've been using the "latest" links as a quick way to see what's happening during a run (or a deployment).

andermi:
1. Fixed
2. would you rather not gzip the log on exit? (only after a duration)

hamilton8415:
Yeah, I think we should match the buoy, zip only at the end of an hour...

andermi:
it was causing problems anyways... if it takes too long to gzip, the batch launch will SIGTERM/SIGKILL pblog before it finishes writing

andermi:
@hamilton8415 ready for another check

hamilton8415:
In testing this I realized the wave-seed wasn't getting passed along by IncidentWave.cpp.  Fixed this and pushed to this branch...

hamilton8415 (approved):
I have used this feature quite a bit in the last week or so and it's working very nicely.  The only changes here I might suggest is to rename the "latest_csv" symbolic link to "latest_csv_dir".  I kept thinking the link pointed at the latest .csv file...  But I got used to it.  Not a critical change.

The other hiccup I've found is maybe not due to this PR but rather the logging PR in _gz. When performing runs longer than one-hour, it zips the last file correctly, but there does seem to be a gap in results between the two files,  when running much faster than real-time the gap is greater than two minutes of sim time...  I will note this in the _gz pull request also.

andermi:
> rename the "latest_csv" symbolic link to "latest_csv_dir"

changed in https://github.com/osrf/mbari_wec_utils/pull/44

andermi:
> gap in results between the two files

fixed in osrf/mbari_wec_utils#44

andermi:
@hamilton8415 can you try your matlab stuff against this latest patch to confirm the new IMU orientation?

hamilton8415:
Lat/Lon working now.

Issue #142, opened 2023-05-02
CI failing to detect 'ros-humble-actuator-msgs'


quarkytale:
All PRs on this repo are stuck on this error, even seeing the same while building the docker image on `mbari_wec`.

```
  apt: command [apt-get install -y ros-humble-actuator-msgs] failed
  apt: Failed to detect successful installation of [ros-humble-actuator-msgs]
Error: Process completed with exit code 1.
```

quarkytale:
> I guess it needs a port to humble? (edited)

Seems like its already added, https://github.com/ros/rosdistro/pull/36795

andermi:
well, somehow it didn't make it into apt after that

andermi:
at the very least, we aren't using anything related to the actuator_msgs stuff so I'm fine reverting and freezing ros_gz for now. We can just tie that commit to our release

andermi:
something like this: https://github.com/osrf/mbari_wec/pull/51

quarkytale:
Adding ros-testing sources list fixes our build, this is a temporary fix till a Humble sync is out (by this week).

andermi:
Please see my comment in #143

quarkytale:
Tested this after Humble sync, apt install worked and could build the latest `ros2` branch of `ros_gz` too!

PR #143, opened 2023-05-02
Temporary fix for CI

quarkytale:


andermi:
I think the key would be `actuator_msgs`

andermi:
try doing a git checkout on ros_gz using commit: 7c62a1ce3cfb4eb3b54e1cfd092c9b63a9d5f255

mabelzhang:
CI is green

mabelzhang:
Thank you both! Could you put a summary back in #142 to say this is temporary until next Humble sync?

andermi:
So, I'm not sure about this fix. I'd rather not release with test deb sources. How soon would we be able to remove this? I think I'd rather freeze the ros_gz source to limit issues like this that sometimes arise from grabbing latest/nightly/test source. We don't really need anything from the latest commit of ros_gz and we will need to freeze it anyways for our release.

cc @hamilton8415

mabelzhang:
We're not shipping ros_gz, right - it is pulled in from rosdep? So to freeze it, we have to manually `git checkout` after the rosdep run? Then that's an additional step for the user. Humble sync should be soon. We could hold our release until Humble sync is out.

mabelzhang:
Well, or we can `git clone` ros_gz to that commit hash in CI and Docker image, instead of letting rosdep do it.

andermi:
ros_gz is already in our vcs yaml file with ros2 branch as the checkout. It's not installed through rosdep

andermi:
my fix was to freeze ros_gz in the yaml which I would like to do anyways for the release.

andermi:
https://github.com/osrf/mbari_wec/pull/51

PR #144, opened 2023-05-10
Log Batch Random Seed

andermi:
if `seed: 0` in batch yaml, generate and log a new random seed per run


hamilton8415 (approved):
This looks good.  At first I was resisting some b/c I didn't like that the seed recorded in batch_results.log was different than in the .yaml file.  But, because the .yaml file is also logged, one can see what's going on and it does provide the possibility of re-creating any given run, which has to be good.

PR #145, opened 2023-05-11
No timeout on scale/retract overrides

andermi:


hamilton8415:
Agreed, there is no timeout that changes the scale factor back to some default.  In the real controller there is actually a scale factor adjust loop that drives the scale up and down based on the standard deviation of the recent RPM, but we haven't used that at sea in a long time and it doesn't need to be simulated here.  Thanks for finding this.

hamilton8415 (approved):

PR #146, opened 2023-05-11
Fix logging rates

andermi:
add sim-time-aware Rate class for controlling publishing in all controllers

### DIFF
--- a/buoy_gazebo/src/IncidentWaves/IncidentWaves.cpp
+++ b/buoy_gazebo/src/IncidentWaves/IncidentWaves.cpp
@@ -74,7 +74,6 @@ void IncidentWaves::Configure(
   gz::sim::EntityComponentManager & _ecm,
   gz::sim::EventManager & /*_eventMgr*/)
 {
-
   this->dataPtr->model = gz::sim::Model(_entity);
   if (!this->dataPtr->model.Valid(_ecm)) {
     ignerr << "IncidentWaves plugin should be attached to a model entity. " <<
--- a/buoy_gazebo/src/buoy_utils/Rate.hpp
+++ b/buoy_gazebo/src/buoy_utils/Rate.hpp
@@ -20,8 +20,6 @@
 
 #include <rclcpp/macros.hpp>
 #include <rclcpp/rate.hpp>
-//#include <rclcpp/utilities.hpp>
-//#include <rclcpp/visibility_control.hpp>
 
 
 namespace buoy_utils
@@ -40,16 +38,17 @@ public:
   : SimRate(
       duration_cast<nanoseconds>(duration<double>(1.0 / rate)),
       _clock
-    )
+  )
   {}
 
-  explicit SimRate(const std::chrono::nanoseconds & period,
-                   rclcpp::Clock::SharedPtr _clock)
+  explicit SimRate(
+    const std::chrono::nanoseconds & period,
+    rclcpp::Clock::SharedPtr _clock)
   : clock(_clock),
     period_(period),
     last_interval_(
       rclcpp::Time(_clock->now(),
-                   _clock->get_clock_type())
+      _clock->get_clock_type())
     )
   {}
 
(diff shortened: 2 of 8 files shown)


andermi:
@hamilton8415 using this branch, I can verify that the pub rates increase with RTF relative to wall, but stay 10Hz relative to sim clock

hamilton8415:
Getting there!  I did some runs with this and it's so nice to be able to run fast, really opens up what's possible...

It looks like perhaps the buoy AHRS (DataID = 3) has been overlooked?  I plotted the time and time-between samples in the log files and everything lines up perfectly on 0.1seconds (when running faster than real-time), but the ^3 lines are still lagging.   These are denoted "XB" in the plot below.

Capture

hamilton8415:
And finally, pbcmd is not exiting gracefully with these pack commands, they work fine though...

hamilton@WOLFPACK:~$ pc_PackRate 25
Executing pc_PackRate to Set the CANBUS packet rate from the Power Controller: 25 Hz
terminate called without an active exception
Aborted (core dumped)

andermi:
I'm pretty certain it comes down to the physics step. If you'd like to have data available every 0.02 seconds, but our physics step is 0.1... We are probably going to have some zero-order-hold going on. I was letting gz-sim decide the timestamp when it updates data, and then the publisher would use that timestamp, but I could hide the ZOH and have the publisher set the stamp when it sends and we'd then have the correct dt in the logs...

andermi:
Can you repeat the test using a physics step of 0.02?

andermi:
As for the pbcmd....  uhhggghghhhghghg

andermi:
TC should be in the same boat as XB but I'm having trouble seeing those datapoints

hamilton8415:
OK, I will check all of this.  Does the RTF take an array of arguments and then run a number of runs?

hamilton8415:
Yes, there is a chance that I ran the physics step slower than 50Hz for this test.  Unfort I deleted the results but will re-run.  makes sense...

> I'm pretty certain it comes down to the physics step. If you'd like to have data available every 0.02 seconds, but our physics step is 0.1... We are probably going to have some zero-order-hold going on. I was letting gz-sim decide the timestamp when it updates data, and then the publisher would use that timestamp, but I could hide the ZOH and have the publisher set the stamp when it sends and we'd then have the correct dt in the logs...



_

andermi:
> Does the RTF take an array of arguments

No, that's one of the globals out of the matrix

hamilton8415:
Here's an updated plot with the physics timestep at 0.01 and a variety of packet rates set via pbcmd.  Still a bit of noise in there, but acceptable.  Our typical matlab processing includes a step in which all of the controllers data points are interpolated onto a common time-base to facilitate analysis.  And that worked fine...

Also, TC data seems fine, just the XB (^3) is not keeping up...

Capture

hamilton8415:
Oh yes, sorry for the red-herring, my bad to ask it to give data more often than it was generating data...

andermi:
the dither there still isn't very satisfying... I'm guessing there's a race between the publish thread grabbing data and the simulation thread updating the data. I wasn't ever too happy with the low-/high-priority triple mutex thing. If it becomes a problem, I could look into a simpler/robust solution.

andermi:
@hamilton8415 can you run your plots for this commit? I think I may have fixed the AHRS pub rate

hamilton8415:
Stand by, running now...

hamilton8415:
Seemingly unchanged for me, but we should probably do a little branch checking.  I'm on fix_logging_rate here and add_pblog_args for _utils...

Also, long buffer times when running RTF>1 present, here on the local machine too...

andermi:
you should be on @rhenthorn 's branch for _utils

andermi:
uhhh oops. I definitely found the problem this time. I never enabled the `use_sim_time` parameter for the AHRS...

andermi:
the fix is in!

hamilton8415 (approved):

PR #147, opened 2023-05-12
Fix pbcmd terminate

andermi:
Needs https://github.com/osrf/mbari_wec_utils/pull/50
fixed the error:

```
terminate called without an active exception
Aborted (core dumped)
```

but still get a segfault every so often with all subcommands of pbcmd

### DIFF
--- a/buoy_gazebo/src/controllers/PowerController/PowerController.cpp
+++ b/buoy_gazebo/src/controllers/PowerController/PowerController.cpp
@@ -34,6 +34,7 @@
 #include <ros_gz_sim/Stopwatch.hpp>
 
 #include <buoy_interfaces/msg/pc_record.hpp>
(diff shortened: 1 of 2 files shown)


quarkytale:
Is this for the batch run seg fault issue?

andermi:
This is to fix an issue with pbcmd not waiting for the request future before quitting and throwing an error. There is also an error that this does not fix which is a segfault in the dds library when pbcmd is closing but only happens about 1 in 10 times (doing up-arrow-enter) while the sim is running.

mabelzhang (approved):
Hope I didn't sign my life away by approving.

andermi:
> Hope I didn't sign my life away by approving.

at least it's not going into main yet ;)

PR #148, opened 2023-05-15
Init latlon for world

andermi:
This warning was being spammed:
```
[Wrn] [Magnetometer.cc:433] Failed to update NavSat sensor enity [84]. Spherical coordinates not set.
```

This PR adds a `spherical_coordinates` to the main world file to init world origin and the warning goes away.

### DIFF
--- a/buoy_gazebo/worlds/buoy_no_plugins.sdf
+++ b/buoy_gazebo/worlds/buoy_no_plugins.sdf
@@ -2,6 +2,14 @@
 <sdf version="1.6">
   <world name="buoy_no_plugins">
 
+    <spherical_coordinates>
+      <surface_model>EARTH_WGS84</surface_model>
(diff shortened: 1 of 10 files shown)


andermi:
@hamilton8415 @quarkytale if one of you could push this one through quickly, I'd appreciate it.

hamilton8415 (approved):
I haven't run and tested this, but looking at the files-changed it seems straightforward and low impact on anything we are doing, so happy to approve.  I will say that I haven't yet looked at the lat/lon reported in the .csv files, other than to notice that they are always zero.   Maybe related?  We will see...

hamilton8415:
OK, I did run this and it runs fine, although with this version of _gz since the PR is against main, the .csv files aren't being created, so we will have to wait to see if lat/lon values appear.  No problem...

PR #150, opened 2023-05-29
Add launch param overrides

andermi:
It became more expedient to test certain run configurations by directly overriding sim params in `mbari_wec.launch.py`

@hamilton8415 I also added the swap for Custom spectra to use freq in Hz in this PR. Incidentally, this is the reason I wanted to be able to override params to test specific configs without batching. I found an instability in the sim for the following config:

```
ros2 launch buoy_gazebo mbari_wec.launch.py inc_wave_spectrum:='inc_wave_spectrum_type:Custom' seed:=42 physics_rtf:=10 physics_step:=0.09 door_state:=closed scale_factor:=0.5 battery_soc:=0.25 x_mean_pos:=0.5
```

It fails almost immediately. If I change to `physics_step:=0.01`, it still crashes but takes longer. And, `physics_step:=0.001` didn't crash in the few minutes I checked, but I can't guarantee it won't still run away eventually.

The buoy & heave cone fly off the screen pretty quickly. 

The crash log:
```
[ruby $(which gz) sim-1] ODE INTERNAL ERROR 1: assertion "aabbBound >= dMinIntExact && aabbBound < dMaxIntExact" failed in collide() [collision_space.cpp:460]
[ruby $(which gz) sim-1] Stack trace (most recent call last):
[ruby $(which gz) sim-1] #31   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df434c96, in
[ruby $(which gz) sim-1] #30   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df431fc5, in
[ruby $(which gz) sim-1] #29   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df42fc34, in
[ruby $(which gz) sim-1] #28   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df37ba1e, in
[ruby $(which gz) sim-1] #27   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df2a69ac, in rb_protect
[ruby $(which gz) sim-1] #26   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df43ec61, in rb_yield
[ruby $(which gz) sim-1] #25   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df43a30c, in rb_vm_exec
[ruby $(which gz) sim-1] #24   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df434c96, in
[ruby $(which gz) sim-1] #23   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df431fc5, in
[ruby $(which gz) sim-1] #22   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df42fc34, in
[ruby $(which gz) sim-1] #21   Object "/usr/lib/x86_64-linux-gnu/ruby/3.0.0/fiddle.so", at 0x7fd3df5c844b, in
[ruby $(which gz) sim-1] #20   Object "/lib/x86_64-linux-gnu/libruby-3.0.so.3.0", at 0x7fd3df3fd088, in rb_nogvl
[ruby $(which gz) sim-1] #19   Object "/usr/lib/x86_64-linux-gnu/ruby/3.0.0/fiddle.so", at 0x7fd3df5c7d6b, in
[ruby $(which gz) sim-1] #18   Object "/lib/x86_64-linux-gnu/libffi.so.8", at 0x7fd3df5b9492, in
[ruby $(which gz) sim-1] #17   Object "/lib/x86_64-linux-gnu/libffi.so.8", at 0x7fd3df5bce2d, in
[ruby $(which gz) sim-1] #16   Object "/usr/lib/x86_64-linux-gnu/libgz-sim7-gz.so.7.5.0", at 0x7fd3df09980d, in runServer
[ruby $(which gz) sim-1] #15   Object "/lib/x86_64-linux-gnu/libgz-sim7.so.7", at 0x7fd3da3397e5, in
[ruby $(which gz) sim-1] #14   Object "/lib/x86_64-linux-gnu/libgz-sim7.so.7", at 0x7fd3da34647a, in gz::sim::v7::SimulationRunner::Run(unsigned long)
[ruby $(which gz) sim-1] #13   Object "/lib/x86_64-linux-gnu/libgz-sim7.so.7", at 0x7fd3da345b40, in gz::sim::v7::SimulationRunner::Step(gz::sim::v7::UpdateInfo const&)
[ruby $(which gz) sim-1] #12   Object "/lib/x86_64-linux-gnu/libgz-sim7.so.7", at 0x7fd3da3423e2, in gz::sim::v7::SimulationRunner::UpdateSystems()
[ruby $(which gz) sim-1] #11   Object "/usr/lib/x86_64-linux-gnu/gz-sim-7/plugins/libgz-sim-physics-system.so", at 0x7fd3a0127a86, in gz::sim::v7::systems::Physics::Update(gz::sim::v7::UpdateInfo const&, gz::sim::v7::EntityComponentManager&)
[ruby $(which gz) sim-1] #10   Object "/usr/lib/x86_64-linux-gnu/gz-sim-7/plugins/libgz-sim-physics-system.so", at 0x7fd3a0128101, in gz::sim::v7::systems::PhysicsPrivate::Step(std::chrono::duration<long, std::ratio<1l, 1000000000l> > const&)
[ruby $(which gz) sim-1] #9    Object "/usr/lib/x86_64-linux-gnu/gz-physics-6/engine-plugins/libgz-physics-dartsim-plugin.so", at 0x7fd35c391a3b, in gz::physics::dartsim::SimulationFeatures::WorldForwardStep(gz::physics::Identity const&, gz::physics::SpecifyData<gz::physics::RequireData<gz::physics::WorldPoses>, gz::physics::ExpectData<gz::physics::ChangedWorldPoses, gz::physics::Contacts, gz::physics::JointPositions> >&, gz::physics::CompositeData&, gz::physics::ExpectData<gz::physics::ApplyExternalForceTorques, gz::physics::ApplyGeneralizedForces, gz::physics::VelocityControlCommands, gz::physics::ServoControlCommands, std::chrono::duration<long, std::ratio<1l, 1000000000l> > > const&)
[ruby $(which gz) sim-1] #8    Object "/lib/x86_64-linux-gnu/libdart.so.6.12", at 0x7fd2d7566e73, in dart::simulation::World::step(bool)
[ruby $(which gz) sim-1] #7    Object "/lib/x86_64-linux-gnu/libdart.so.6.12", at 0x7fd2d754c705, in dart::constraint::ConstraintSolver::solve()
[ruby $(which gz) sim-1] #6    Object "/lib/x86_64-linux-gnu/libdart.so.6.12", at 0x7fd2d754afe1, in dart::constraint::ConstraintSolver::updateConstraints()
[ruby $(which gz) sim-1] #5    Object "/lib/x86_64-linux-gnu/libdart-collision-ode.so.6.12", at 0x7fd3629761cb, in dart::collision::OdeCollisionDetector::collide(dart::collision::CollisionGroup*, dart::collision::CollisionOption const&, dart::collision::CollisionResult*)
[ruby $(which gz) sim-1] #4    Object "/lib/x86_64-linux-gnu/libode.so.8", at 0x7fd35c51a266, in dxHashSpace::collide(void*, void (*)(void*, dxGeom*, dxGeom*))
[ruby $(which gz) sim-1] #3    Object "/lib/x86_64-linux-gnu/libode.so.8", at 0x7fd35c5227b7, in dDebug
[ruby $(which gz) sim-1] #2    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fd3dee287f2, in abort
[ruby $(which gz) sim-1] #1    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fd3dee42475, in raise
[ruby $(which gz) sim-1] #0    Object "/lib/x86_64-linux-gnu/libc.so.6", at 0x7fd3dee96a7c, in pthread_kill
[ruby $(which gz) sim-1] Aborted (Signal sent by tkill() 951633 11883)
```

### DIFF
--- a/.github/workflows/build-and-test.sh
+++ b/.github/workflows/build-and-test.sh
@@ -62,8 +62,8 @@ colcon build --packages-up-to buoy_tests --event-handlers console_direct+
 source $COLCON_WS/install/setup.bash
 
 # Test all buoy packages
-#colcon test --packages-select-regex=buoy --packages-skip=buoy_msgs --event-handlers console_direct+ --retest-until-pass 10
-#colcon test-result
+colcon test --packages-select-regex=buoy --packages-skip=buoy_msgs --event-handlers console_direct+ --retest-until-pass 10
+colcon test-result
 
 # Debug specific test
-launch_test $COLCON_WS/install/buoy_tests/share/buoy_tests/launch/pc_bias_damping_ros_feedback_py.launch.py
+# launch_test $COLCON_WS/install/buoy_tests/share/buoy_tests/launch/pc_bias_damping_ros_feedback_py.launch.py
--- a/.github/workflows/build-and-test.sh
+++ b/.github/workflows/build-and-test.sh
@@ -62,5 +62,8 @@ colcon build --packages-up-to buoy_tests --event-handlers console_direct+
 source $COLCON_WS/install/setup.bash
 
 # Test all buoy packages
-colcon test --packages-select-regex=buoy --packages-skip=buoy_msgs --event-handlers console_direct+ --retest-until-pass 10
-colcon test-result
+#colcon test --packages-select-regex=buoy --packages-skip=buoy_msgs --event-handlers console_direct+ --retest-until-pass 10
+#colcon test-result
+
+# Debug specific test
+launch_test $COLCON_WS/install/buoy_tests/share/buoy_tests/launch/pc_bias_damping_ros_feedback_py.launch.py
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -84,7 +84,7 @@ else:
     inc_wave_spectrum = lambda : '<!-- No Waves -->'
 
 ############################
-# Mean Piston Position
+# Piston Position
 ############################
 
 # Would like to specify piston mean position and set all other
@@ -94,12 +94,19 @@ else:
 # that brings the equilibrium to the desired mean piston position.
 # Recompute pressure and volume based on Ideal Gas Law.
 
+# Check if initial piston position (m) was passed in via empy
+try:
+    initial_piston_position
+except NameError:
+    initial_piston_position = 0.7  # m, default;
+                                   # measured from fully retracted (same as range_finder)
+
 # Check if x_mean_pos (m, mean piston position) was passed in via empy
-# with the current parameters, the mean position settles around 1.21m
 try:
     x_mean_pos
 except NameError:
     x_mean_pos = 0.7  # m, default; measured from fully retracted (same as range_finder)
+
 
 import math
 import scipy.optimize as spo
@@ -195,6 +202,7 @@ if not ignore_piston_mean_pos:
       <JointName>HydraulicRam</JointName>
       <chamber>upper_polytropic</chamber>
       <is_upper>true</is_upper>
+      <initial_piston_position>@(initial_piston_position)</initial_piston_position>
       <!-- measure of valve opening cross-section and duration (meter-seconds) -->
       <valve_absement>7.77e-6</valve_absement>
       <pump_absement>4.3e-8</pump_absement>
--- a/buoy_description/models/mbari_wec/model.sdf.em
+++ b/buoy_description/models/mbari_wec/model.sdf.em
@@ -43,15 +43,15 @@ def bretschneider_spectrum(Hs=2.0, Tp=13.0):
       <Tp>{Tp}</Tp>
 ''')
 
-def custom_spectrum(w=None, Szz=None):
+def custom_spectrum(f=None, Szz=None):
     ''' Prints the Custom <IncWaveSpectrumType> block for the IncidentWave plugin. '''
-    if w is None:
-        w = [0.0, 0.2, 0.4, 0.6, 2.0]
+    if f is None:
+        f = [0.0, 0.2, 0.4, 0.6, 2.0]
     if Szz is None:
         Szz = [0.0, 0.4, 1.0, 1.0, 0.0]
     coefs = \
-        ('\n' + 6*' ').join([f'<w{idx}>{wn}</w{idx}> <Szz{idx}>{Szzn}</Szz{idx}>'
-            for idx, (wn, Szzn) in enumerate(zip(w, Szz))])
+        ('\n' + 6*' ').join([f'<f{idx}>{fn}</f{idx}> <Szz{idx}>{Szzn}</Szz{idx}>'
+            for idx, (fn, Szzn) in enumerate(zip(f, Szz))])
     print(f'''
       <IncWaveSpectrumType>Custom</IncWaveSpectrumType>
       {coefs}
@@ -77,7 +77,7 @@ elif 'Bretschneider' in inc_wave_spectrum_type:
         inc_wave_spectrum = bretschneider_spectrum  # default
 elif 'Custom' in inc_wave_spectrum_type:
     try:
-        inc_wave_spectrum = partial(custom_spectrum, w, Szz)
+        inc_wave_spectrum = partial(custom_spectrum, f, Szz)
     except NameError:
         inc_wave_spectrum = custom_spectrum  # default
 else:
--- a/buoy_description/models/mbari_wec_base/model.sdf.em
+++ b/buoy_description/models/mbari_wec_base/model.sdf.em
@@ -96,10 +96,10 @@ pto_inner_radius = tether_radius + pto_gap
 pto_scale = pto_inner_radius / pto_stl_inner_radius
 
 # PTO Buoyancy (specify desired COB, computes location of extra volume above .stl collision mesh)
-pto_disp = .205  # m^3  (Specified)
-pto_cob = -3  # z-position, on centerline (Specified)
-stl_disp = .160443 # m^3  Specific to .stl file
-stl_cob = 0 # z-position, on centerline. Buoyancy plugin uses collisioin pose origin, not center of mesh volume
+pto_disp = 0.205  # m^3  (Specified)
+pto_cob = -3.0  # z-position, on centerline (Specified)
+stl_disp = 0.160443 # m^3  Specific to .stl file
+stl_cob = 0.0 # z-position, on centerline. Buoyancy plugin uses collisioin pose origin, not center of mesh volume
 buoyancy_disp = pto_disp - stl_disp
 buoyancy_cob = (pto_cob*pto_disp - stl_cob*stl_disp)/buoyancy_disp
 buoyancy_radius = ((3*buoyancy_disp)/(4*math.pi))**(1/3)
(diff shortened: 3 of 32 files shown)


hamilton8415:
Couple of things.   Changing this to Hz requires the latest fs_hydrodynamics library, which I haven't quite get put on the ppa server.  I will do that.  Perhaps you're compiling that library though, and in that case I have to think about what the default "custom" spectra might be, that could be nonsense causing a really fast input wave...

andermi:
At the moment, without the new library, it is only a variable name change (no actual math changed to Hz instead of rad/s)

andermi:
We should probably figure out how to protect the sim from crashing in this manner with a loud warning that some values are getting clipped

hamilton8415:
Oh yes, of course, just a name change...  OK, but what spectra are you supplying with the custom designation. Perhaps the default being used in gazebo (or in the library) is screwy.  I will try to re-create...

andermi:
default is

```
w = [0.0, 0.2, 0.4, 0.6, 2.0]
Szz = [0.0, 0.4, 1.0, 1.0, 0.0]
```

andermi:
I also noticed the return code from the crash was not being captured by the `LaunchService` so I added some stuff to get the return code directly and properly log it to `batch_runs.log`. So, now we can actually see that the sim crashed on those runs.

andermi:
I decreased testing time on buoy_tests down to ~35% of what it used to be, and it's still timing out in CI on github... We may need to look into using a faster self-hosted / cloud runner or pay for upgraded github runners. This is getting painful.

andermi:
@mjcarroll @mabelzhang no matter what I do I cannot get the future to be ready here:
https://github.com/osrf/mbari_wec_gz/blob/3b97b6b8855918880d462075eae880253ecf7a0d/buoy_tests/testing_utils/utils.py#L327-L333

called from the `pc_bias_damping_ros_feedback_py.launch.py` test in `buoy_tests`:
https://github.com/osrf/mbari_wec_gz/blob/3b97b6b8855918880d462075eae880253ecf7a0d/buoy_tests/launch/pc_bias_damping_ros_feedback_py.launch.py#L70

for the service here:
https://github.com/osrf/mbari_wec_gz/blob/3b97b6b8855918880d462075eae880253ecf7a0d/buoy_tests/tests/fixture_server_sinusoidal_piston.cpp#L96-L129

The issue is only on the slow github CI. I've tried my trick of rate limiting the executor spin among various other things to no avail. The service call shows complete but the future isn't updated and the await blocks forever. It is repeatable in CI but I can't get it to happen on my machine.

mjcarroll:
> @mjcarroll @mabelzhang no matter what I do I cannot get the future to be ready here:

Looking

andermi:
osrf/mbari_wec_utils#51 should fix test issue

andermi:
woot! tests pass in 28 minutes now!

andermi:
@hamilton8415 a few good changes and bugfixes in this PR\. I kept the original buoy height and piston position for tests and set mbari_wec.launch.py to use a better initial piston position and buoy height by default

hamilton8415:
Here's a comparison between a Bretschneider spectrum (Hs = 2m, Tp = 10s) with what we have been using as the default "custom" spectrum.  Presumably the crashes are occuring due to the excess high-frequency content, which would make some sense that a smaller time-step stabilizes things some. 
Capture

hamilton8415:
Here's a candidate spectrum for a default custom spectrum, just a sub-sampling of the Bretschider spectrum from above:
w = [0.007, 0.164, 0.321, 0.479, 0.634, 0.793, 0.948, 1.106, 1.262, 1.419, 1.576, 1.736];  // rad/sec
Szz = [ 0,  0.000, 0.000, 0.760, 2.278, 1.518,  0.800, 0.414,  0.225, 0.129,  0.078,  0.0484];  m^2/(rad/sec)

hamilton8415:
And of course, once we switch to using the latest fs_hydro library which expects these spectrum relative to Hertz, these values will need to become:
f = [0.0011,  0.0262, 0.0511, 0.0762, 0.1009, 0.1262, 0.1509, 0.1760, 0.2009, 0.2259, 0.2508, 0.2762];  // Hz
Szz = [ 0, 0.0000, 0.0000, 4.7783, 14.3138, 9.5359, 5.0249, 2.6016, 1.4153, 0.8100, 0.4880, 0.3043];  // m^2/(Hz)

andermi:
those values do better, but the sim still crashes after 45 mins (sim time) using RTF 10 and step 0.01

hamilton8415:
Hmm, but it runs longer with bretscneider 2m/10s?  If that’s true it has to do with the smoothness of the supplied spectrum, which is surprising, and likely a bug…
    	
    	Get Outlook for iOS
    
  
 From: andermi ***@***.***>Sent: Friday, June 2, 2023 12:58 PMTo: osrf/mbari_wec_gzCc: Andrew Hamilton; MentionSubject: Re: [osrf/mbari_wec_gz] Add launch param overrides (PR #150) 
those values do better, but the sim still crashes after 45 mins (sim time) using RTF 10 and step 0.01

andermi:
bretschneider went for over an hour without crashing

andermi:
@hamilton8415 try this with osrf/mbari_wec_utils#52

hamilton8415:
With regard to the issue of a custom spectrum being selected without a spectrum being specified, I have worked on this to try and identify why this is, it seems that it's related to the sparseness of the custom spectra specifications (i.e. spectrum specified at only a few frequencies).  I have investigated this and haven't found anything obvious, and I don't have an idea of a physical or numerical reason this would matter.   

I think we could do two things, one is to specify the default custom spectrum used at a higher resolution and see if the crash still occurs.  Or, since it's a bit non-nonsensical anyway to specify a custom spectrum and not list what that spectrum is, we could have it default to using the internally generated 1m/10s Bretschnieder spectrum in the case that "custom" is specified without a spectrum description.

In either case, we should not this as an issue/bug somewhere as it's sure to turn back up...

andermi:
Well, I think the issue is that a user could specify a spectrum that crashes the sim. Is there a way we could detect unreasonable forces in the wave plugin and cap them with a warning that the spectrum doesn't have enough resolution or even error out and close the sim?

hamilton8415:
Yes, agreed.  I will do some more work on this to try and figure out what's going on.  Without that understanding, it's hard to check for...  Even if I could find a characterization of the spectrum that fails reliably, we could at least warn/prohibit that...  Or I suppose a pre-interpolation, but that doesn't make sense that would change anything...  Interpolate now, interpolate later...

hamilton8415 (approved):
For me, everything is working as expected.  Let's merge this into main and address the sensitivity to the custom wave-spectrum specification elsewhere.

PR #152, opened 2023-07-12
incoming wave height service

andermi:
Adds ROS 2 service to check incoming wave height and normal vector (slope of wave) at an array of points.

requires https://github.com/osrf/mbari_wec_utils/pull/53 and https://github.com/hamilton8415/FreeSurfaceHydrodynamics/pull/5

NOTE: had to change the LinearIncidentWave object in ECM to non-pointer... need to get to the bottom of that if we want to keep it as a pointer.

for testing:
[send_iwh_req.py.zip](https://github.com/osrf/mbari_wec_gz/files/12533940/send_iwh_req.py.zip)


andermi:
@hamilton8415 I'm adding a stamp to evaluate the height at a specific time, and was wondering if I should make it relative to current time or absolute? I feel like relative to current time would be more useful... e.g. evaluate 10 seconds from now

andermi:
@hamilton8415 I made it relative for now (easily changed)

hamilton8415:
Only hiccup would be if there's a question of when "now" is, presumably it will return relative to the sim time when the service request is handled, that may be slightly different than the time in user-code, which would be when the last ROS messages were sent out.  Small differences though, using absolute time may alleviate that. May, I need to think more...

andermi:
In the response, I'm having it return the requested relative_time value as well as the simulation time it was computed

andermi:
So, a user could see when exactly that height is valid using stamp + relative_tine

andermi:
> series of times into the future (5-20 seconds worth), if the simulation advances while these requests are being made

I had envisioned a single request for all rather than a series of requests. I could use an array for the relative_time to be 1-to-1 per point so it would be a snapshot and no advancement would happen... I just figured this would be less bookkeeping for the user. Absolute time would require them to still do the math on their end anyways?

For your second item (and setting up coords at-sea), there's a couple ways to do it...

1. all bodies in lat/lon and Vincenty's formula for relative dist/az which could be converted to cartesian
2. all bodies in local cartesian about fixed world lat/lon, but we don't have an SBL system so this would just be (1) with the cartesian conversion relative to the fixed world lat/lon rather than between bodies

hamilton8415:
Ah yes, if an array of times at a single location is possible, that works fine then.  I'm on the fence, so let's leave it relative for now.

I think option (2) for the locations, stick to cartessian, with a specified anchor location as the origin, that's certainly easiest in the simulator, and if outside wave-buoys report in lat/lon, then conversion relative to a fixed anchor point might be needed.

andermi:
> I'm on the fence

When in doubt, do both! :-p

updated the python script:
[send_iwh_req.py.zip](https://github.com/osrf/mbari_wec_gz/files/12540198/send_iwh_req.py.zip)

andermi:
still working on (2)

hamilton8415:
Of course I knew you would do both. :)

andermi:
done with (2)

hamilton8415:
This is all working pretty nicely for me.  I am a bit unsure how to access the absolute versus relative time request I thought you were hinting at.  The lastest .py file didn't have an obvious option.  Similarly, does that tool multiple location/time requests, and if so what's the syntax.

andermi:
oh, I can update the tool to do absolute_time... I just hardcoded relative

andermi:
latest python tool
[send_iwh_req.py.zip](https://github.com/osrf/mbari_wec_gz/files/12543629/send_iwh_req.py.zip)

you can add multiple points by just making space-separated values after -x, -y, -abs_t/-rel_t

```
$ ./send_iwh_req.py -h
usage: send_iwh_req.py [-h] [-x X [X ...]] [-y Y [Y ...]] [--use_buoy_origin]
                       [--rel_t REL_T [REL_T ...]] [--abs_t ABS_T [ABS_T ...]]

options:
  -h, --help            show this help message and exit
  -x X [X ...]          one or more x points to evaluate wave height
  -y Y [Y ...]          one or more y points to evaluate wave height
  --use_buoy_origin     points are relative to buoy origin (world otherwise)
  --rel_t REL_T [REL_T ...]
                        enter one or more relative time (+/-) values in decimal seconds from sim
                        time to compute height
  --abs_t ABS_T [ABS_T ...]
                        enter one or more absolute time values in decimal seconds to compute height
```

`$ ./send_iwh_req.py -x 1 2 3 -y 4 5 6 --rel_t 0.0 0.0 0.0 --use_buoy_origin`

hamilton8415:
Got it.  All features are evident to me now.  Looks great, the flexibility here should cover all of the cases I can imagine.  Ship it. :)

hamilton8415 (approved):

PR #158, opened 2023-08-28
add matrix_mode: False option to sim params yaml to allow vectorized batch params

andermi:
`matrix_mode: False` for vectorized batch params instead of meshgrid


hamilton8415 (approved):
I haven't tested this but it seems we should move forward with getting these merged into to main and we can then test them all together.  I see tests are failing, possibly because of the versioning issue with FreeSurfaceHydrodynamics.

hamilton8415:
Yes, I'm catching up with the roadmap!

PR #160, opened 2023-09-05
Debug flag for WaveBodyInteractions plugin

andermi:
Right now, if we want to see debug statements from another plugin, WaveBodyInteractions produces so much output, that it is difficult to see other prints. Also, logging easily fills up disk space in `~/.gz/sim/log`. So, move gzdbg messages for the WaveBodyInteractions plugin behind debug flag.


hamilton8415 (approved):
This suits me fine.  An alternative would be to comment out or even remove most of these debug messages.  They were there for initial debug during development.  Under the presumption that things are working as expected it'd be fine to remove them.  There is a day when someone/I goes to add features related to multiple wave directions or similar, but they/I could add a lot of this back in which would probably be a useful an necessary exercise in understanding some of the details.  This addition is a low priority...

PR #163, opened 2023-09-19
Fix logging rates

andermi:
increased executor spin loop rate for all controllers

see also osrf/mbari_wec_utils#56


hamilton8415 (approved):
Part Deux branch and https://github.com/osrf/mbari_wec_utils/pull/56 work fine for me.  .csv data outputs are at 10Hz, running at desired RTF of 100.

PR #164, opened 2023-09-19
Fix mbari_wec.launch.py arguments

andermi:
override arguments weren't displayed in `-s` option of launch file
typo in inc wave override description


hamilton8415 (approved):
Works for me, -s shows options.  Tested with experiment_comparison_select.launch.py

But, I remember it working yesterday too, so consider the source. :)

PR #165, opened 2023-09-19
Add wave body total power to latent data

andermi:
requires osrf/mbari_wec_utils#57

### DIFF
--- a/buoy_gazebo/src/LatentData/LatentData/LatentData.hpp
+++ b/buoy_gazebo/src/LatentData/LatentData/LatentData.hpp
@@ -150,20 +150,26 @@ struct WaveBody
 
   gz::math::Vector3d buoyant_force{0.0, 0.0, 0.0};
   gz::math::Vector3d buoyant_moment{0.0, 0.0, 0.0};
+  double buoyancy_total_power{0.0};
   gz::math::Vector3d radiation_force{0.0, 0.0, 0.0};
   gz::math::Vector3d radiation_moment{0.0, 0.0, 0.0};
+  double radiation_total_power{0.0};
   gz::math::Vector3d exciting_force{0.0, 0.0, 0.0};
   gz::math::Vector3d exciting_moment{0.0, 0.0, 0.0};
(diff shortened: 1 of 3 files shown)


hamilton8415:
Looks good, it's hard to verify that the powers are correct, and one can see that the leading order term of the excitation force is the opposite of the buoyancy force, as it should be in this linear theory.  Makes it a bit hard for either one to have meaning without the other, but I think we should leave it this way as that's usually how these things are represented.  

We may want to add the position and velocity of the buoy to this message, i.e. w_Pose_p for the position, and w_xdot_p and w_omega_p for the velocities.  I think that could be useful information for a number of things.

Also, since we're here, it'd be great to include the local incident wave height that's being used to compute the exciting force.  Looks like I did some trickery so that the exciting force computation figures out what that is on it's own, so there isn't an explicit assignment.  Looking at that it's computed as " _IncWave->eta(0, 0, t);" in the wave-exciting force code, and we have that incident wave object so should be easy to get it.  Note that it has to be called at x=0,y=0.  This isn't very satisfying, but how it has to be, it basically means the origin of the wave-field is moving around with the buoy, and a point to discuss related to WecSIM, I am not sure how they handle the issue but it could be related to the slight phase difference we see when comparing results.

Anyway, I am also happy to add these last features if desired.

hamilton8415:
Right, and immediately after hitting send, I remember that we already have the capability to show the wave-height data.   Showing that at the buoy origin should do the trick, so maybe we can leave it at that.

andermi:
I added
pose: position, orientation
twist: linear/angular velocities

andermi:
and like you said, we have the wave height at the buoy position being reported in latent data by default

hamilton8415:
I don't think anything is neccessary, that location is where the buoyancy force balances the weight.  If it's off, there's no big effect other than in calm water there's a steady-state buoy dispaclement to account for the difference.

andermi:
I think this is ready to go then

hamilton8415 (approved):
This runs and looks good to me.

PR #167, opened 2023-09-21
Update CI

andermi:
I goofed...


hamilton8415 (approved):
That was bound to happen sooner or later...

PR #168, opened 2023-12-05
Change WAMIT data to mbari_snl version

hamilton8415:
This is a simple change from one set of buoy hydrodynamic coefficients for the buoy to an updated set that was computed by the Sandia group.  This new data is contained in files called mbari_snl_IR.1, mbari_snl_JR.2, mbari_snl.1, and mbari_snl_3.  The WAMIT output was supplied in an .h5 file (included but not used in this repository), and converted to the .1/.3 data format by auxiliary scripts.


andermi (approved):
looks good! One thought would be to put the `auxiliary scripts` to convert he `.h5` in the repo, but perhaps we can put them with our other scripts in the matlab/python processing repo we were discussing...

hamilton8415:
Yes, I considered including that.  In my typical fashion it's "results oriented", but it needs to be archived somewhere so I'll put it in and maybe add some notes to the beginning to explain how it could be improved.

PR #189, opened 2026-03-03
Add seed batching

alexje76:
Added the capability to batch across seeds

### DIFF
--- a/buoy_gazebo/scripts/mbari_wec_batch
+++ b/buoy_gazebo/scripts/mbari_wec_batch
@@ -150,18 +150,6 @@ def generate_simulations(sim_params_yaml):
                                 ' must be a single value')
         sys.exit(-1)
 
-    # seed (random seed for incident waves plugin)
-    try:
-        if 'seed' in sim_params:
-            seed_ = int(sim_params['seed'])
-        else:
-            seed_ = None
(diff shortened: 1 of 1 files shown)


alexje76:
This addition has been working well for me. I attempted to match all formatting and naming conventions, but there may be some followed that I did not pick up on

buoy_gazebo/scripts/mbari_wec_batch
```diff
@@ -214,6 +202,18 @@ def generate_simulations(sim_params_yaml):
...
+        if 'seed' in sim_params:
+            seeds = np.array(sim_params['seed'], dtype=int)
+            seeds = np.atleast_1d(seeds)
+        else:
+            seeds = None
+            node.get_logger().warn('sim_params_yaml: optional seed parameter not specified')
+    except TypeError:
+        node.get_logger().error('sim_params_yaml: seed parameter' +
+                                ' must be a single value')
+        sys.exit(-1)
```
andermi: Since we are setting seeds to be an array, we don't need this error anymore.

andermi (changes requested):
Other than that one `TypeError` that we don't need anymore, this looks great! Thanks for adding it :)

andermi (approved):

PR #190, opened 2026-07-15
Seed enumeration

alexje76:


andermi:
@alexje76 were you wanting to make the rosbag optional?

alexje76:
I had been yes - there was an issue where I ran out of memory while running inside a container (apptainer not docker) on the UW HPC cluster, though it ended up unclear if this was the root cause of the issue there, I continued to run it on that system successfully

andermi (approved):
I tweaked the rosbag2 arg a little, but good to go!

